PR #3, opened 2018-05-15
Add dummy FlatList for ActivityFeed

tinaroh:


src/shared/ActivityFeed.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { FlatList, View, Text, StyleSheet } from "react-native";
+
+type ListProps = {};
```
osdiab: I'd just call this `ActivityFeedProps` to match convention with other stuff

PR #4, opened 2018-05-18
Add Redux Store, Firebase Auth flow

osdiab:
Adds tidied up redux store from raha-web-app, and Google and Facebook auth to the app.

login-mobile


src/data/prod.config.json
```diff
@@ -0,0 +1,24 @@
+{
```
rahulgi: Do we want to use TS config files instead? I like that model since it gives us the property checking.

src/store/reducers/operations.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export type Operation = OperationBase &
+  (
+    | {
+        op_code: OperationType.REQUEST_INVITE;
+        data: RequestInvitePayload;
+      }
+    | {
+        op_code: OperationType.TRUST;
```
rahulgi: You don't want to include GIVE / MINT operations yet?

osdiab: Yeah I copied it before you made those I can copy those in before I merge

src/data/prod.config.json
osdiab: True will do

PR #5, opened 2018-05-18
Onboarding Flow

osdiab:
Add basic swiper for onboarding flow.

appvideo


src/components/pages/Onboarding.tsx
```diff
@@ -8,18 +9,33 @@ type OnboardingProps = {
 export default class Onboarding extends React.Component<OnboardingProps> {
   render() {
     return (
-      <View style={styles.container}>
-        <Text>This is the onboardibng page.</Text>
+      <View style={[styles.slide, { backgroundColor: "#C04DEE" }]}>
+        <Text style={styles.header}>Give Raha</Text>
+        <Text style={styles.text}>
+          By using Raha, you are supporting basic income for everyone.
```
markulrich: What's the copy for swipe pages? How about
1. Raha is a payments network where transaction fees fund a basic income for all members.
1. To create a trusted network and prevent fraud and theft, Raha is invite only and all members must verify identity via video.
1. Create your video today to join Raha and join an economy where we can help each other and recognize every life has value.
Also I like what "Shop your Way" does where you can click/swipe next or just hit "Get Started" to skip over

tinaroh:
Updated with the following:
- Added back the Sign Up button for now
- Added the pages with some initial copy. I think we can discuss the final copy separately
- Commented out the Android special-casing in Swiper for now since I didn't know how to resolve the TypeScript errors
- Add config copy step to README

Do you want me to finish out the onboarding flow completely in this PR or can I address in multiple other ones? I think maybe we should merge this one since there are a few things that can be worked on concurrently:
1. make the Swiper styling take in column of elements properly, it turns out everything is absolutely positioned so adding new buttons doesn't work currently. Want to fix this before doing what Mark proposed
2. Detect when user is a new user and take them into onboarding and remove Sign Up button
3. Continue through onboarding flow using the camera after JOIN NOW is clicked
4. Figure out why the Android special-casing is needed, uncomment, and fix TypeScript

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
+/**
```
osdiab: General note on this, it does seem like this is a good deal more general and flexible than what we need in our own codebase (the number of properties you can pass it is quite expansive and technical!) but given that it works let's just keep it; saying it with an eye to keep the complexity of our codebase down. Perhaps we can just expose a subset of this configurability through a higher level, simpler component.

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+  StyleSheet,
+  View,
+  ViewPagerAndroidOnPageScrollEventData
+} from "react-native";
+import RoundedButton from "./RoundedButton";
+
+// Detect screen width and height
+const { width, height } = Dimensions.get("window");
+
+type SwiperProps = {
```
osdiab: I don't know what a few of these actually do, some  comments would be nice if we do keep all this configurability.

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /**
+   * Initialize the state
+   */
+  initState(props: SwiperProps): SwiperState {
+    // Get the total number of slides passed as children
+    const total = props.children ? props.children.length || 1 : 0,
+      // Current index
+      index =
```
osdiab: Would prefer if we did this as two separate const statements for clarity, like

```
const total = ...;
const index = ...;
const offset = ...;
```

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  /*
+     * Drag end handler
+     */
+  onScrollEndDrag: ScrollViewProps["onScrollEndDrag"] = event => {
+    if (!event) {
+      // TODO: what does this case even mean? read react-native docs
+      return;
+    }
+    const {
```
osdiab: Similar here about just doing these as separate const calls, it just looks odd to me lol

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+    this.setState({
+      index
+    });
+  };
+
+  /**
+   * Swipe one slide forward
+   */
+  swipe = () => {
+    // Ignore if already scrolling or if there is less than 2 slides
```
osdiab: This comment is likely to become out of date since it just describes the code, I think no need for it.

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * Render pagination indicators
+   */
+  renderPagination = () => {
+    if (this.state.total <= 1) {
+      return null;
+    }
+
+    const ActiveDot = <View style={[styles.dot, styles.activeDot]} />,
```
osdiab: I think cloning these elements is kinda strange React code, I'd probably do it this way:

```
renderPagination = () => {
  if (this.state.total <= 1) {
    return null;
  }

  // this weird Array crap is just a way to generate a range from `0` to `(this.state.total - 1)` inclusive, maybe we 
  // can put that in a helper module
  const dots = [...Array(this.state.total)].keys()].map(index => {
    const styles = [styles.dot, ...(index === this.state.total ? [styles.activeDot] : [])];
    return <View style={styles} />
  });

  return <View pointerEvents="none" style={[styles.pagination, styles.fullScreen]}>{dots}</View>
}
```

src/components/shared/Swiper.tsx
```diff
@@ -0,0 +1,329 @@
...
+  container: {
+    backgroundColor: "transparent",
+    position: "relative"
+  },
+  slide: {
+    backgroundColor: "transparent"
+  },
+  pagination: {
+    position: "absolute",
+    bottom: 110,
```
osdiab: I wonder if we really need to do this kind of absolute positioning, but it's not important enough right now to really be a big deal

src/components/shared/Swiper.tsx
tinaroh: Will do!

src/components/shared/Swiper.tsx
tinaroh: Nope, it doesn't need to be. Added the issue here https://github.com/rahafoundation/raha-mobile-app/issues/14

PR #18, opened 2018-05-30
Redux persistence

osdiab:
Persists the Redux store on the device. Also retrieves it upon load.

Designed so that when we move the redux store to its own library, the react-native specific persistence stuff can be kept in this repo.

If https://github.com/rt2zz/redux-persist/pull/834 gets merged, then the typings will work; otherwise I'll have to remove them.

## TODO

- [x] add the react bindings for redux-persist
- [x] ensure it works properly
- [x] figure out how this impacts dev flow
  - We should probably add a button to sign off
  - [And maybe a way to purge the stored state on dev as well](https://github.com/rt2zz/redux-persist/issues/579)

Will not land until the above is fixed, as it is not important for this to land immediately.


App.tsx
```diff
@@ -25,11 +26,14 @@ const Navigator = createStackNavigator(
...
 );
 
-const store = createStore();
+const { store, persistor } = createStore();
 const App: React.StatelessComponent<{}> = () => {
   return (
     <Provider store={store}>
-      <Navigator />
+      {/* TODO: decide if we should show a loading indicator here. */}
+      <PersistGate loading={null} persistor={persistor}>
```
tinaroh: Is it standard for this to be null? I'd imagine this to be a boolean or optional.

osdiab: That's from the docs and the typings don't seem to be complaining, so guess that's how this library does it lol. In React any falsey value won't be rendered so shouldn't matter if we passed in undefined or what not.

PR #20, opened 2018-06-06
Add OnboardingCamera page which shows camera preview and prompt text

tinaroh:
- Removed `react-native-camera` in favor of Expo Camera which doesn't require us to eject the apps yet
- Gets CAMERA and AUDIO permissions from user
- Fixed overstretched Home page on Android due to `minHeight: 100%` (it might just be my Galaxy S8 though). `flex: 1` seems to be enough to stretch the content.

Up next: Finish recording should advance user to a preview screen where they can upload.

*Other Issues:*
- Denying permission seems to crash Expo though, looking into that separately since stacktrace points outside our app
- I was trying to use [Camera.Constants.Type.back](https://github.com/expo/expo-sdk/blob/master/src/Camera.js#L90) as suggested [here](https://docs.expo.io/versions/latest/sdk/camera#type) but the type definition doesn't include it so I hardcoded the constants "front" and "back" for now. Any ideas on how to resolve?

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -9,7 +9,7 @@ import Home from "./src/components/pages/Home";
 import LogIn from "./src/components/pages/LogIn";
 import OnboardingCamera from "./src/components/pages/OnboardingCamera";
 import Onboarding from "./src/components/pages/Onboarding";
-import createStore, { store, persistor } from "./src/store";
+import { store, persistor } from "./src/store";
 import { refreshMembers } from "./src/store/actions/members";
 
 export enum RouteName {
--- a/App.tsx
+++ b/App.tsx
@@ -7,6 +7,7 @@ import { AsyncStorage } from "react-native";
 
 import Home from "./src/components/pages/Home";
 import LogIn from "./src/components/pages/LogIn";
+import OnboardingCamera from "./src/components/pages/OnboardingCamera";
 import Onboarding from "./src/components/pages/Onboarding";
 import createStore, { store, persistor } from "./src/store";
 import { refreshMembers } from "./src/store/actions/members";
@@ -21,6 +22,9 @@ const Navigator = createStackNavigator(
   {
     Home: {
       screen: Home
+    },
+    OnboardingCamera: {
+      screen: OnboardingCamera
     },
     Onboarding: {
       screen: Onboarding
--- a/App.tsx
+++ b/App.tsx
@@ -15,6 +15,7 @@ import { refreshMembers } from "./src/store/actions/members";
 export enum RouteName {
   Home = "Home",
   Onboarding = "Onboarding",
+  OnboardingCamera = "OnboardingCamera",
   LogIn = "LogIn"
 }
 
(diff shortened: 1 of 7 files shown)


osdiab:
with respect to the screen extending below the controls, @rahulgi  was experiencing that too on his android device so it's not just you. Perhaps the parent height actually includes the entire screen, not just the area above the navigation controls, so `min-height: 100%` would use the space below the controls; that's my best guess assuming the CSS-ish styling they use in `react-native` works like how web CSS works, so `flexbox` style stuff would just expand to the space available, rather than the actual space of the parent. In short, sounds good to me.

osdiab:
With response to the typings, you can make a PR to add them :) If they're defined in [the DefinitelyTyped monorepo](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/expo), which has some good guidelines on how to contribute—I've made a bunch of revisions to library types before and I think it's a good way to better understand how TypeScript works as well.

App.tsx
```diff
@@ -7,13 +7,15 @@ import { AsyncStorage } from "react-native";
 
 import Home from "./src/components/pages/Home";
 import LogIn from "./src/components/pages/LogIn";
+import OnboardingCamera from "./src/components/pages/OnboardingCamera";
 import Onboarding from "./src/components/pages/Onboarding";
-import createStore, { store, persistor } from "./src/store";
```
osdiab: Sorry for that, I'll add some CI testing in a bit. I disabled the lint checks for unused imports before, but it would make sense if we turned that on again I think.

src/components/pages/OnboardingCamera.tsx
```diff
@@ -0,0 +1,36 @@
...
+  render() {
+    return (
+      <React.Fragment>
+        <Camera
+          onVideoRecorded={uri => {
+            console.log("Video URI: " + uri);
+          }}
+        />
+        <Text style={styles.text}>
+          Example: "My name is Jane Doe and I'm joining Raha because I believe
```
osdiab: Looks like there are some whitespace issues? If you have `prettier` installed in your VSCode and autoformatting enabled it'll take care of this stuff for you automatically (in this case it'd replace the trailing space with `{" "}` since the linter prefers no trailing spaces anywhere.

src/components/shared/Camera.tsx
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  renderFlipButton = () => {
+    return (
+      <TouchableOpacity
+        style={styles.flipButton}
+        onPress={() => {
+          this.setState({
+            type: this.state.type === "back" ? "front" : "back"
```
osdiab: Maybe leave a quick TODO for posterity here that once the expo types include the constants we ought to use them here.

src/components/pages/OnboardingCamera.tsx
tinaroh: I do have prettier installed. Can you clarify which trailing space you mean? I don't think there's a trailing space anywhere. Not sure what this GitHub highlighting means.

tinaroh: Looking into it, GitHub apparently thinks there's an error here based on https://github.com/github/linguist/tree/master/vendor but it's a false positive since the code runs fine

PR #25, opened 2018-06-17
Home page feed

osdiab:
## TODO
- [x] Load true operations to memory
- [x] Style operations as something more readable than JSON
- [x] Implement each operation type
- [x] Make `ActivityFeed` filterable


src/components/shared/ActivityFeed.tsx
```diff
@@ -1,36 +1,77 @@
 import * as React from "react";
```
tinaroh: Can we leave quick comments that describe each component at the top?

osdiab: donezoes

PR #31, opened 2018-06-18
Add onMemberSelected callback for MemberSearchBar

tinaroh:
When a user selects a member from the search bar, it will return the Member object to the passed in callback.

### DIFF
--- a/src/components/shared/MemberSearchBar.tsx
+++ b/src/components/shared/MemberSearchBar.tsx
@@ -2,7 +2,13 @@
  * Component that allows search through members and suggests autocompletions.
  */
 import * as React from "react";
-import { FlatList, View, StyleSheet, Text } from "react-native";
+import {
+  FlatList,
+  View,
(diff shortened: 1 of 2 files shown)


src/components/shared/MemberSearchBar.tsx
```diff
@@ -75,11 +83,19 @@ class MemberSearchBarView extends React.Component<
           onClearText={() => this.clearSuggestions()}
         />
         <FlatList
+          keyboardShouldPersistTaps="always"
```
tinaroh: FYI this is for the issue listed here: https://github.com/facebook/react-native/issues/4229  -- the first tap on the item was dismissing the soft keyboard and the second tap would do the onPress. This property makes it so that the first tap goes to the onPress

rahulgi: could you leave add this as a comment in code?

PR #32, opened 2018-06-19
Navigation tabs

markulrich:


src/components/shared/Navigation.tsx
```diff
@@ -0,0 +1,180 @@
...
+            break;
+          case RouteName.Invite:
+            iconName = "account-multiple-plus";
+            break;
+          case RouteName.Search:
+            iconName = "ios-search";
+            IconType = Ionicons;
+            break;
+          default:
+            iconName = "bad";
```
osdiab: What does `bad` mean? If this is supposed to never happen, I say put a `console.error` call, a `TODO` comment to remind us to log it properly in the future then return nothing (like an empty `<View />`)

src/components/shared/Navigation.tsx
```diff
@@ -0,0 +1,180 @@
...
+            IconType = Ionicons;
+            break;
+        }
+        // You can return any component that you like here! We usually use an
```
osdiab: Remove comment?

src/components/shared/Navigation.tsx
```diff
@@ -0,0 +1,180 @@
...
+      screen: MyProfile
+    }
+  } as { [key in RouteName]: any }, // TODO: once react-nav types in, edit
+  {
+    initialRouteName: RouteName.MyProfile,
+    labeled: false,
+    navigationOptions: ({ navigation }: any) => ({
+      tabBarIcon: ({ focused }: any) => {
+        const { routeName } = navigation.state;
+        let iconName = null;
```
osdiab: You don't need to initialize it

src/components/shared/Navigation.tsx
```diff
@@ -0,0 +1,180 @@
...
+  OnboardingCamera = "OnboardingCamera",
+  LogIn = "LogIn",
+  MyProfile = "MyProfile",
+  VideoPreview = "VideoPreview",
+  Search = "Search",
+  Invite = "Invite",
+  Give = "Give"
+}
+
+const SignedIn = createMaterialBottomTabNavigator(
```
osdiab: `SignedInNavigator` (and same for `SignedOut` would be clearer

PR #44, opened 2018-06-19
Add Give page.

rahulgi:


src/components/pages/Give/GiveForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+type StateProps = {
+  loggedInMember?: Member;
+  getMemberById: (memberId: MemberId) => Member | undefined;
+  apiCallStatus: ApiCallStatus | undefined;
+};
+type DispatchProps = {
+  give: typeof give;
+};
+type MergedProps = StateProps & {
+  give: (amount: string, memo: string, toMemberId: MemberId) => void;
```
osdiab: Can we just pass around `Big` instead of using `string` in some places and `Big` elsewhere? I think we should only convert to and from `Big` at the endpoints, i.e. when we're writing to and from UI, across the network, or to and from DB. That'd make it easier to have consistent APIs all over our code and also ensure minimal loss of precision, etc

rahulgi:
screenshot_20180619-163654

rahulgi:
screenshot_20180619-163717

src/components/pages/Give/GiveForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+  toMember?: Member;
+} & FormState;
+
+class GiveFormView extends React.Component<Props, State> {
+  public constructor(props: Props) {
+    super(props);
+    const toMemberId = props.toMemberId;
+    this.state = {
+      toMember: toMemberId ? props.getMemberById(toMemberId) : undefined,
+      amount: "",
```
osdiab: So this would be probably undefined until a Big value gets set

src/components/pages/Give/GiveForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+import { getMembersByIds } from "../../../store/selectors/members";
+import { MemberSearchBar } from "../../shared/MemberSearchBar";
+
+const MESSAGE_PLACEHOLDER_TEXT = "";
+const MAX_MEMO_LENGTH = 140;
+
+type OwnProps = {
+  toMemberId?: MemberId;
+  identifier: string;
+  onSuccessCallback: (toMember: Member, amount: string, memo: string) => any;
```
osdiab: Please leave a doc comment explaining what `identifier` and `onSuccessCallback` are, as it's not immediately apparent

src/components/pages/Give/GiveForm.tsx
```diff
@@ -0,0 +1,328 @@
...
+                excludeMembers={
+                  this.props.loggedInMember ? [this.props.loggedInMember] : []
+                }
+              />
+            </View>
+          )}
+        </View>
+        {this.state.toMember ? (
+          <Text style={styles.section}>
+            {this.state.toMember.fullName} is currently donating 3% of all Raha
```
osdiab: Is the donation rate dynamically stored somewhere? Either way leave a comment.

src/components/pages/Give/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  toMember?: Member;
+  amount?: string;
+  memo?: string;
+};
+
+/**
+ * Return a generally unique identifier.
+ */
+function getOperationIdentifer(): string {
+  return new Date().getTime().toString();
```
osdiab: good enough for this I guess lol. Would be a bit safer if you append some random nonce at the end probably

src/components/pages/Give/GiveForm.tsx
```diff
@@ -0,0 +1,318 @@
...
+    );
+  };
+  private onChangeAmount = (amount: string) => {
+    if (amount.length === 0) {
+      this.setState({ amount: undefined });
+    } else {
+      try {
+        // Rounding Mode half-up
+        const bigAmount = new Big(amount).round(2, 1);
+        if (this.validateAmount(bigAmount)) {
```
osdiab: It looks like this function isn't complete?

PR #48, opened 2018-06-20
Create Mint page

markulrich:


markulrich:
screenshot_20180620-145433

src/shared/IosAndroidSafeAreaView.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {
+  Text,
+  View,
+  SafeAreaView,
+  ViewProps,
+  Platform,
+  StatusBar
+} from "react-native";
+
+export const IosAndroidSafeAreaView: React.StatelessComponent<
```
osdiab: Can we just call it `SafeAreaView`? or `SaferAreaView` lol. Don't like the platform-specific naming.

src/components/pages/Mint.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <IosAndroidSafeAreaView style={{ flex: 1 }}>
+      <View
+        style={[
+          styles.centerFlex,
+          { flexDirection: "row", marginHorizontal: 50 }
+        ]}
+      >
+        <View style={styles.centerFlex}>
+          <Text style={styles.number}>
+            {loggedInMember.balance.minus(30).toString()}
```
osdiab: Can you leave a comment saying that this is a placeholder lol.

osdiab:
I think it's not really intuitive as it stands but we can think about copy and the design language later

PR #50, opened 2018-06-21
Refactor Onboarding to be controlled by a single parent component

tinaroh:
Onboarding component will own the onboarding subcomponents and the redux state.


src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -0,0 +1,184 @@
...
+            retakeVideo={(errorMessage?: string) => {
+              this.setState({
+                errorMessage: errorMessage,
+                step: OnboardingStep.CAMERA
+              });
+            }}
+          />
+        );
+      }
+      default:
```
osdiab: Not necessary?

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -0,0 +1,184 @@
...
+            step: OnboardingStep.CAMERA
+          });
+          break;
+        }
+        return (
+          <VideoPreview
+            videoUri={videoUri}
+            retakeVideo={(errorMessage?: string) => {
+              this.setState({
+                errorMessage: errorMessage,
```
osdiab: Can just type it once, shorthand will do what you want

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -0,0 +1,184 @@
...
+                step: OnboardingStep.CAMERA
+              });
+            }}
+          />
+        );
+      }
+      case OnboardingStep.CAMERA: {
+        // Shouldn't happen, but if any required field is cleared or the onboarding
+        // flow is screwed up, redirect to the correct step.
+        const verifiedFullName = this.state.verifiedName;
```
osdiab: Maybe make these verification things be a helper, this switch statements is getting pretty long

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+type OnboardingState = {
+  step: OnboardingStep;
+  invitingMember?: Member;
+  verifiedName?: string;
+  videoUri?: string;
+  errorMessage?: string;
+};
+
+export class OnboardingView extends React.Component<
```
osdiab: It just occurred to me, a shortcoming of this approach is that native back button on android for example wouldn’t work... if we care about that working then maybe the approach would be to do the page switching with a nested StackNavigator or something from React-navigation instead of via a switch statement. Still would be able to do this prop passing like it is here, but that way the nav native bindings should work?

tinaroh: Will just close this PR out and keep the existing SignedOut StackNavigator.

PR #56, opened 2018-06-21
Implement video upload to firebase during onboarding.

tinaroh:
Renamed the component to InviteVideoPreview too since it caused some confusion earlier. I didn't make it an action since as discussed, the video upload doesn't affect the state of any other part of the app.


src/components/pages/Onboarding/InviteVideoPreview.tsx
```diff
@@ -0,0 +1,243 @@
...
+    });
+  }
+
+  uploadVideo = async (videoUploadRef: firebase.storage.Reference) => {
+    const videoUri = this.props.navigation.getParam("videoUri", null);
+    this.setState({ uploadStatus: UploadStatus.UPLOADING });
+    const response = await fetch(videoUri);
+    const blob = await response.blob();
+    //@ts-ignore Expo Blob does not have data type
+    if (blob.data.size <= MAX_VIDEO_SIZE) {
```
osdiab: To reduce nesting, I’d do 

```
if (blob.data.size > MAX_VIDEO_SIZE) {
  This.setState({ // ...error })
  Return
}

// do the rest here

File: src/components/pages/Onboarding/InviteVideoPreview.tsx [hunk 2]
```diff
@@ -0,0 +1,243 @@
...
+        errorMessage:
+          "The video size is larger than " +
+          MAX_VIDEO_SIZE +
+          "MB. Please retake your video."
+      });
+    }
+  };
+
+  requestInvite() {
+    this.setState({
```
osdiab: I think you can use @rahulgi ’s code for `Give` as a model for doing this, ask him about it

PR #60, opened 2018-06-22
Switch param name to username.

rahulgi:


src/api/index.ts
```diff
@@ -53,7 +53,7 @@ type GetOperationsApiCall = ApiCallDefinition<
 type RequestInviteApiCall = ApiCallDefinition<
   ApiEndpoint.REQUEST_INVITE,
   { memberId: MemberId },
-  { fullName: string; videoUrl: string; creatorMid: string }
+  { fullName: string; videoUrl: string; username: string }
```
tinaroh: Hm, maybe we can do creatorUsername? Or make the other one inviterMemberId

PR #61, opened 2018-06-22
Request invite after video upload completes

tinaroh:


src/components/pages/Onboarding/InviteVideoPreview.tsx
```diff
@@ -115,17 +130,26 @@ class InviteVideoPreviewView extends React.Component<
           errorMessage: "Could not upload. Please try again.",
           uploadStatus: UploadStatus.NOT_STARTED
         }),
-      () => {
+      async () => {
         this.setState({ uploadStatus: UploadStatus.UPLOADED });
-        this.requestInvite();
+        const videoDownloadUrl = await uploadTask.snapshot.ref.getDownloadURL();
+        if (videoDownloadUrl) {
+          this.sendInviteRequest(videoDownloadUrl);
```
markulrich: Should we wait to set it to `UPLOADED` here and do `UPLOADING` before?

tinaroh: Yeah!

PR #74, opened 2018-06-24
Add Discover tab

markulrich:


src/components/shared/Navigation.tsx
```diff
@@ -184,7 +208,7 @@ const SignedOutNavigator = createStackNavigator(
     InviteVideoPreview: {
       screen: InviteVideoPreview
     }
-  } as { [key in RouteName]: any }, // TODO: once react-nav types in, edit
+  },
```
osdiab: Why remove this?

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+  title: string | string[];
+  footer?: string | string[];
+  uri: string;
+};
+
+type DiscoverCard = {
+  header?: string;
+  title: string;
+  footer?: string;
+  uri: (navigation: NavigationScreenProp<{}>) => void;
```
osdiab: this is a weird name for the actual value

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export const DiscoverWebView: React.StatelessComponent = ({
+  navigation
+}: any) => {
+  return <WebViewNative source={{ uri: navigation.getParam("uri") }} />;
+};
+
+type DiscoverCardRaw = {
+  header?: string | string[];
+  title: string | string[];
```
osdiab: probably should be body

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+  footer?: string | string[];
+  uri: string;
```
osdiab: this seems a little strict for now since not every card will necessarily be pointing to a remote source, but ok

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+    return (navigation: NavigationScreenProp<{}>) => {
+      navigation.navigate(RouteName.DiscoverWebView, { uri });
+    };
+  }
+  if (uri.startsWith("mailto:")) {
+    return (navigation: NavigationScreenProp<{}>) => {
+      Linking.openURL(uri); // TODO check canOpen
+    };
+  }
+  throw Error(`Invalid uri ${uri}, unsupported protocol`);
```
osdiab: If we're throwing an error we probably should be careful about not crashing the app by calling this - so try catch around whatever calls this, for now it can just console.error and then return some empty value

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+}
+
+function convertCard(discoverCard: DiscoverCardRaw): DiscoverCard {
```
osdiab: Why does this need to be a reduce call if there's only 4 fields? Why not just

```typescript
function pickRandomFromArr(arr?: any[]) {
  if (!arr || arr.length === 0) {
    return undefined;
  }
  return arr[Math.floor(arr.length * Math.random())];
}

function convertCard(discoverCard: DiscoverCardRaw): DiscoverCard {
  return {
    header: pickRandomFromArr(discoverCard.headerOptions),
    body: pickRandomFromArr(discoverCard.bodyOptions),
    footer: pickRandomFromArr(discoverCard.footerOption),
    uri: convertUriToCallback(discoverCard.uri)
  }
}
```

src/components/pages/Discover.tsx
```diff
@@ -0,0 +1,179 @@
...
+        }
+      }
+      res[key as keyof DiscoverCard] = val;
+      return res;
+    },
+    {} as DiscoverCard
+  );
+}
+
+function convertCardArr(cardArr: DiscoverCardRaw[]): DiscoverCard[] {
```
osdiab: This seems a little silly to be a function but shrug

PR #85, opened 2018-06-28
Add standard display components.

rahulgi:
Adds standard display components from which levers like font can be controlled.

The components are: `Button`, `Text`, and `Container`. Check them out at `src/components/display`. The rest of this PR is adding custom fonts and using these custom display components throughout the app.


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
+    "link-assets": "react-native link",
```
osdiab: This also links libraries and stuff, but if this makes it easier for you then whatevs - that said react-native link is a very commonly referenced in libraries, and simple command so not sure if it really ought to be added to scripts

src/components/display/Text.tsx
```diff
@@ -0,0 +1,23 @@
+/**
```
osdiab: Why not put these in shared as well? Not sure what the display subdir communicates

src/shared/SafeAreaView.tsx
```diff
@@ -22,7 +21,7 @@ export const SafeAreaView: React.StatelessComponent<ViewProps> = props => {
   const { children, ...remaining } = props;
   return (
     <View {...remaining}>
-      <View style={{ height: StatusBar.currentHeight }} />
+      {/* <View style={{ height: StatusBar.currentHeight }} /> */}
```
osdiab: You can probably just delete thus

PR #93, opened 2018-06-29
Style give form.

rahulgi:


rahulgi:
screenshot_20180629-153211

screenshot_20180629-153304

osdiab:
I think the text inputs don't have enough visual distinction to clearly communicate that they're text inputs, and I also think it'd look clearer if there were things that had more visual weight, like bolded amounts of Raha, donation amount, and person names. I'm gonna find some examples and post them here

src/components/pages/Give/GiveForm.tsx
```diff
@@ -228,6 +240,9 @@ class GiveFormView extends React.Component<Props, State> {
               (this.props.apiCallStatus &&
                 this.props.apiCallStatus.status === ApiCallStatusType.STARTED)
             }
+            buttonStyle={{ backgroundColor: "#2196F3" }}
```
osdiab: Doesn't have to happen now but we should gather the disparate colors we use and place them into a palette file like in the web repo, as they're already getting repeated and copy pasted within this PR.

To go into further detail, probably describe them first as things like "accent color, background color, etc..." and then define in the button how they would be mapped to functional uses of those colors (like add a prop to Button that specifies it's nature: "primary button" maps to the accent color, "secondary button" maps to some shade down of the accent color, "danger button" maps to a red thing or something, etc).

osdiab:
here's some ideas that are super fancy, i don't like all of them since sometimes the color contrast is too low and the animations are generally superfluous, but the point is pretty much all of them are clearly things you can interact with, not just text. 

https://medium.com/swlh/creative-form-input-field-design-examples-bfe5dd50808a

osdiab:
screen shot 2018-06-29 at 5 03 42 pm
screen shot 2018-06-29 at 5 04 15 pm
screen shot 2018-06-29 at 5 07 19 pm
screen shot 2018-06-29 at 5 07 30 pm
screen shot 2018-06-29 at 5 07 35 pm
screen shot 2018-06-29 at 5 07 54 pm

rahulgi:
Good feedback, I'm gonna merge for now since I think it looks better than what's currently there and let's iterate and improve over time.

PR #95, opened 2018-07-03
Add Invite flow component and upload video

tinaroh:
Uploads invite video to "invite-video" bucket in Firestore which prevents `list` queries but allows `get` from authenticated users. I did it through Firestore since it had precedence -- could look into uploading directly to Google Cloud Storage later.

Uses a randomly generated alphanumeric string as the token which will be forwarded in the SEND_INVITE flow through the URL send in the email which will serve as the deeplink for the recipient.  

TODO before merge: 
- [ ] Update Raha prod Firestore rules

### DIFF
--- a/src/components/pages/Invite/Invite.tsx
+++ b/src/components/pages/Invite/Invite.tsx
@@ -7,11 +7,11 @@ import { VideoPreview } from "../Camera/VideoPreview";
 import { MapStateToProps, connect } from "react-redux";
 import {
   getLoggedInMember,
-  getPrivateVideoInviteRef
+  getInviteVideoRef
 } from "../../../store/selectors/authentication";
 import { Member } from "../../../store/reducers/members";
 import { RahaState } from "../../../store";
-import { v4 as uuid } from "uuid";
+import { generateToken } from "../../../helpers/token";
 
 enum InviteStep {
   CAMERA,
@@ -23,9 +23,7 @@ type ReduxStateProps = {
   loggedInMember?: Member;
 };
 
-type OwnProps = {
-  videoUploadRef: firebase.storage.Reference;
-};
+type OwnProps = {};
 
 type InviteProps = ReduxStateProps & OwnProps;
 
@@ -37,12 +35,14 @@ type InviteState = {
 };
 
 export class InviteView extends React.Component<InviteProps, InviteState> {
-  inviteId: string;
+  inviteToken: string;
+  videoUploadRef: firebase.storage.Reference;
   dropdown: any;
 
   constructor(props: InviteProps) {
     super(props);
-    this.inviteId = uuid();
+    this.inviteToken = generateToken();
+    this.videoUploadRef = getInviteVideoRef(this.inviteToken);
     this.state = {
       step: InviteStep.CAMERA
     };
@@ -107,7 +107,7 @@ export class InviteView extends React.Component<InviteProps, InviteState> {
         return (
           <VideoPreview
             videoUri={videoUri}
-            videoUploadRef={this.props.videoUploadRef}
+            videoUploadRef={this.videoUploadRef}
             onVideoUploaded={(videoDownloadUrl: string) =>
               this.setState({
                 videoDownloadUrl: videoDownloadUrl
--- a/src/components/pages/Invite/Invite.tsx
+++ b/src/components/pages/Invite/Invite.tsx
@@ -1,0 +1,158 @@
+import { BackHandler } from "react-native";
+import * as React from "react";
+import { Container } from "../../shared/elements";
+import { InviteCamera } from "./InviteCamera";
+import DropdownAlert from "react-native-dropdownalert";
+import { VideoPreview } from "../Camera/VideoPreview";
+import { MapStateToProps, connect } from "react-redux";
+import {
+  getLoggedInMember,
+  getPrivateVideoInviteRef
+} from "../../../store/selectors/authentication";
+import { Member } from "../../../store/reducers/members";
+import { RahaState } from "../../../store";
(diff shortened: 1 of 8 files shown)


src/components/pages/Invite/Invite.tsx
```diff
@@ -0,0 +1,158 @@
...
+    super(props);
+    this.inviteToken = generateToken();
+    this.videoUploadRef = getInviteVideoRef(this.inviteToken);
+    this.state = {
+      step: InviteStep.CAMERA
+    };
+  }
+
+  componentDidMount() {
+    BackHandler.addEventListener("hardwareBackPress", this._handleBackPress);
```
osdiab: Cool!

src/components/pages/Invite/Invite.tsx
```diff
@@ -0,0 +1,158 @@
...
+                step: InviteStep.VIDEO_PREVIEW
+              });
+            }}
+          />
+        );
+      }
+      case InviteStep.VIDEO_PREVIEW: {
+        const videoUri = this._verifyVideoUri();
+        if (!videoUri) {
+          return <React.Fragment />;
```
osdiab: Maybe log something as well? if this is unexpected

src/components/pages/Invite/Invite.tsx
```diff
@@ -0,0 +1,158 @@
...
+                errorMessage
+              );
+              this.setState({
+                step: InviteStep.CAMERA
+              });
+            }}
+          />
+        );
+      }
+      default:
```
osdiab: And also log an error here or below.

src/components/pages/Invite/InviteCamera.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const mapStateToProps: MapStateToProps<
+  ReduxStateProps,
+  OwnProps,
+  RahaState
+> = state => {
+  const loggedInMember = getLoggedInMember(state);
+  // Should never be undefined if you're never logged in, but if you are magically, you get to be Midoriya!
+  return {
+    ownFullName: loggedInMember ? loggedInMember.fullName : "Izuku Midoriya"
```
osdiab: Haha excellent. Though I think what we also ought to log something if it ever was that state as well.

src/helpers/token.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Generates an alphanumeric string token of length 10 or 11.
+ */
+export function generateToken() {
+  return Math.random()
```
osdiab: I think this should be fine since this is just for generating an obscure uri, but it's worth being aware that this random generator is not cryptographically secure.

src/components/pages/Invite/Invite.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+type ReduxStateProps = {
+  loggedInMember?: Member;
+};
+
+type OwnProps = {};
+
+type InviteProps = ReduxStateProps & OwnProps;
+
+type InviteState = {
```
osdiab: Not necessary and potentially overcomplex, but you can potentially make this bullet-proof typewise by defining the states that are allowed explicitly:

```typescript
type BaseInviteState = {
  step: InviteStep
};

// or whatever it should be, when is `verifiedName` not present?
type CameraStepInviteState = BaseInviteState & {
  verifiedName: string
};
type VideoPreviewStepInviteState = CameraStepInviteState & {
 videoUri: string;
}
type CompleteInviteState = VideoPreviewStepInviteState & {
  videoDownloadUrl: string;
}
type InviteState = CameraStepInviteState | VideoPreviewStepInviteState | CompleteInviteState;
```

osdiab: edited it a bit to be even more precise

src/components/pages/Invite/Invite.tsx
tinaroh: Apparently logging can lag the app and in general I don't think we want console logs in our prod apps. I think we should log analytics when we have it. Otherwise I think the error dropdown should be enough to alert us during debugging.

src/helpers/token.ts
tinaroh: Yeah, I'm doing this knowing that it's not cryptographically secure, but I don't think we need a cryptographically secure URI here?

src/components/pages/Invite/InviteCamera.tsx
tinaroh: I think this is fine, since it's not application critical

src/components/pages/Invite/Invite.tsx
tinaroh: Same as above, not sure what console.error gains us

src/components/pages/Invite/Invite.tsx
osdiab: I just want at least something here so that it's obvious that this is a place we should be logging. I don't want to have to audit all the code to figure out all the places we might have missed that logs should occur; so at a minimum I think we should put a comment and even better would just to install that babel plugin and include it in our babelrc.

src/components/pages/Invite/InviteCamera.tsx
osdiab: I want to know when our application breaks expectations we think should never occur, so I would like to have something here to note it

src/helpers/token.ts
osdiab: don't think so either

src/components/pages/Invite/Invite.tsx
tinaroh: Sure. I'll install the plugin in a follow up PR.

PR #106, opened 2018-07-12
Add send invite page for email

tinaroh:
Includes basic email validation. API has not yet been changed yet server-side so the API call isn't tested yet, but figured I can check in the page beforehand.

### DIFF
--- a/src/components/pages/Invite/SendInvite.tsx
+++ b/src/components/pages/Invite/SendInvite.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { Button, TextInput } from "react-native";
+import { Button, TextInput, View } from "react-native";
 import { connect, MapStateToProps, MergeProps } from "react-redux";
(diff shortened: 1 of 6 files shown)


src/components/pages/Invite/SendInvite.tsx
```diff
@@ -0,0 +1,107 @@
...
+        return <Text>Invite failed.</Text>;
+      default:
+        return undefined;
+    }
+  };
+
+  render() {
+    return (
+      <React.Fragment>
+        <TextInput
```
osdiab: I'd make it a [controlled component](https://reactjs.org/docs/forms.html#controlled-components) for state consistency, but I don't think trimming will cause any inconsistency issues so it's fine as is

tinaroh: I'll look into it with the additional validation stuff!

PR #116, opened 2018-07-18
Enable multidex for Android

tinaroh:
Android will not build if we hit > 64k methods. See: https://developer.android.com/studio/build/multidex

I hit this limit on my branch importing a library. Separating this out just in case you guys hit it.

For prod, we should enable Proguard to strip out unused methods/resources. Added an issue for it.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -103,6 +103,7 @@ android {
         targetSdkVersion 22
         versionCode 1
         versionName "1.0"
+        multiDexEnabled true
         ndk {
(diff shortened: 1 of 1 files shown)


android/app/build.gradle
```diff
@@ -151,6 +152,9 @@ android {
 }
 
 dependencies {
+    // Multidex support for < Android 5.0 (API level 21)
```
rahulgi: Should this be greater than instead of less than?

tinaroh: Nope, this library is not required for newer versions.

PR #117, opened 2018-07-18
Transcode videos to remove video metadata and make size smaller.

tinaroh:
Videos are transcoded to 480x640 and bitrate is reduced by 3x.

The original library used is https://github.com/shahen94/react-native-video-processing -- and while the master branch is active, the maintainer hasn't pushed a new release, so I imported a forked version (https://github.com/tinaroh/react-native-video-processing). I also hit a crash in compress which I have a PR out for and have the fix in the fork.

Also fixed a bug with Onboarding redirecting back to VERIFY_NAME when it should go to REQUEST_INVITE which I accidentally committed.

Apparently there isn't a standard with video metadata, but I checked the resulting video's metadata with https://www.get-metadata.com and verified Android tags present in the original version are now gone.


android/app/build.gradle
```diff
@@ -151,6 +152,9 @@ android {
 }
 
 dependencies {
+    // Multidex support for < Android 5.0 (API level 21)
```
rahulgi: looks like this PR also includes the Multidex PR #116?

tinaroh: Yeah sorry, didn't split it out properly. I'll rebase and remove this

src/components/pages/Camera/VideoPreview.tsx
```diff
@@ -51,8 +53,31 @@ export class VideoPreview extends React.Component<
...
-  uploadVideo = async (videoUploadRef: firebase.storage.Reference) => {
-    const response = await fetch(this.props.videoUri);
+  compressAndUploadVideo = async () => {
+    const options = {
+      width: 480,
+      height: 640,
+      bitrateMultiplier: 3,
+      minimumBitrate: 300000
+    };
+    this.videoPlayerRef
```
osdiab: You can do this async await style

try {
  const newSource = await this.videoPlayerRef.compress(options);
  // bleh
catch (error) {
  // blah
}

tinaroh: Am I supposed to mix try/catches with Promises? What do I return from error flow?

osdiab: in an `async` function, when a promise you `await` on results in an error in a `try` block, then it triggers the `catch` block; and when an `async` function throws an exception, it results in the Promise rejecting rather than doing a normal exception. 

So in fact it's better that way! the try/catch block will handle both normal exceptions and rejected promises, and throwing promises from an `async` function will do the expected behavior of rejecting the promise.

PR #121, opened 2018-07-19
Fix temporary Invite splash page

tinaroh:
We should probably stop the users before they go through the Camera flow. I previously put this on the last page

### DIFF
--- a/src/components/pages/Invite/Invite.tsx
+++ b/src/components/pages/Invite/Invite.tsx
@@ -19,7 +19,7 @@ import { SendInvite } from "./SendInvite";
 const ENABLE_SEND_INVITE = false;
 
 enum InviteStep {
-  WIP,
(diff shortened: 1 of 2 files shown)


src/components/pages/Invite/Invite.tsx
```diff
@@ -16,7 +16,10 @@ import { RahaState } from "../../../store";
 import { generateToken } from "../../../helpers/token";
 import { SendInvite } from "./SendInvite";
 
+const ENABLE_SEND_INVITE = false;
+
 enum InviteStep {
+  WIP,
```
osdiab: What is the meaning of the WIP step? Work in Progress doesn't really make sense to me

tinaroh: It means that Invite flow is WIP. It's temporary anyway

PR #123, opened 2018-07-19
HockeyApp

osdiab:
Still haven't gotten it to install properly on my phone, may have to do with signing. Will inform when ready


rahulgi:
such app much hockey

osdiab:
Stale, I'm going to close this for now until a time that we actually decide to use this.

Issue #76, opened 2018-06-25
Deeplink into app


tinaroh:

osdiab:
Is this complete? @tinaroh

tinaroh:
No, I need to incorporate redirecting people to SMS login before onboarding and HTTPS deeplinks

tinaroh:
Implemented but not enabled until HTTPS/HTTP is complete https://github.com/rahafoundation/raha-mobile-app/issues/119

PR #126, opened 2018-07-20
Add LogIn for Onboarding via deeplink

tinaroh:
If the user reaches Onboarding without being logged in, that means they navigated directly via deeplinking. Ask them to log in before continuing.

#76


src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -198,6 +201,17 @@ export class OnboardingView extends React.Component<
   };
 
   _renderOnboardingStep() {
+    if (!this.props.isLoggedIn) {
+      return (
+        <React.Fragment>
+          <Text style={{ textAlign: "center" }}>
+            Welcome to Raha! Please log in to accept your invite.
```
rahulgi: Should this be Log in or Sign up?

tinaroh: hmmm I guess sign up!

PR #131, opened 2018-07-21
Accept https://*.raha.app links as deep links

tinaroh:


tinaroh:
Please re-review! @osdiab or @rahulgi 

react-navigation wasn't forwarding the HTTPS links so I did it manually.

PR #143, opened 2018-07-24
only set confirmation code timer on mount

osdiab:
i can't run on android, and even if i did I can't trigger firebase to auto continue log in because my phone number is on my iphone not my sim card. @rahulgi can you let me know if you still see the error related to confirmation code interval?

### DIFF
--- a/src/components/pages/LogIn.tsx
+++ b/src/components/pages/LogIn.tsx
@@ -46,6 +46,7 @@ import { Map } from "immutable";
 import { TextInput } from "../shared/elements/TextInput";
 import { fonts } from "../../helpers/fonts";
 import { colors } from "../../helpers/colors";
+import { Loading } from "../shared/Loading";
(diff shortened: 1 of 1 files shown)


src/components/pages/LogIn.tsx
```diff
@@ -238,10 +238,16 @@ class ConfirmationCodeForm extends React.Component<
   };
 
   componentDidMount() {
+    this.setState({
+      timerInterval: setInterval(this._calculateTimeLeft, 1000)
```
markulrich: see https://discuss.reactjs.org/t/setinterval-in-state/7955, `setState` usually triggers a render which isn't needed here.

Also reminder about line 451 `new Date()`

osdiab:
@rahulgi @markulrich one of you pl0x briefly recheck this stale, refurbished PR

PR #150, opened 2018-07-26
Add inactivity timer, vote, and account recovery info to Account

markulrich:


src/components/pages/Account.tsx
```diff
@@ -20,32 +28,69 @@ type DispatchProps = {
 
 type StateProps = {
   loggedInMember: Member;
+  trustedForRecovery: Member;
+  votingFor: Member;
+};
+
+type Props = DispatchProps & OwnProps & StateProps;
```
tinaroh: Not sure how important this is, but I think we usually do AccountProps / AccountState?

src/components/pages/Account.tsx
```diff
@@ -20,32 +28,69 @@ type DispatchProps = {
...
+
+type State = {
+  timeRemaining: string;
 };
 
-type Props = DispatchProps & OwnProps;
+const Break: React.StatelessComponent<{}> = () => <View style={styles.break} />;
 
 // TODO show when account will go inactive, total Raha outstanding,
```
tinaroh: Can you remove this now?

src/components/pages/Account.tsx
```diff
@@ -20,32 +28,69 @@ type DispatchProps = {
...
+    };
+  }
+
+  private getTimeRemaining() {
+    // TODO make a RED countdown with finer granularity than 1 day
+    // if you're within a month of being marked inactive, call
+    // updateTimeRemaining on setInterval.
+    const sinceLastMintMilli =
+      Date.now() - this.props.loggedInMember.lastMinted.getTime();
+    return `${400 - Math.round(sinceLastMintMilli / MS_PER_DAY)} days`;
```
tinaroh: Shouldn't the timer be since last transaction instead of last minted?

src/components/pages/Account.tsx
```diff
@@ -20,32 +28,69 @@ type DispatchProps = {
...
+  break: {
+    backgroundColor: colors.border2,
+    height: 5,
+    marginTop: 2,
+    marginBottom: 2
   },
-  barText: {
-    fontSize: 20
+  timeRemaining: {
+    fontWeight: "800"
```
tinaroh: This doesn't apply to both Android/iOS equally I think. In fonts.ts we don't actually support 800 for any of our fonts. Maybe use 'normal', 'bold' instead?

src/components/shared/VideoWithPlaceholder.tsx
```diff
@@ -18,20 +18,18 @@ import {
...
-
 const initialState = {
   isPressed: false,
   videoLoaded: false,
   videoPaused: true,
 };
 
 export class VideoWithPlaceholderView extends React.Component<
-  Props,
+  Props & NavigationInjectedProps,
```
tinaroh: What was wrong with what you were doing before?

src/store/selectors/members.ts
```diff
@@ -39,13 +39,21 @@ export function getMembersByUsernames(
   );
 }
 
+/**
+ * @param member Return all ancestors of this member.
+ * @param byMemberId Map from member id to member.
+ * @return A 2-element array of the set of all ancestors and ordered array of all ancestors.
+ * Everyone is their own ancestor, so given a valid member this is garunteed to contain at least one.
```
tinaroh: nit: sp garunteed

src/components/pages/Account.tsx
markulrich: Grep shows 14 instances of just `Props`, can we make standard be that if file has multiple `Props` then disambiguate all of them, but if just one okay to leave as `Props`?

src/components/pages/Account.tsx
markulrich: Good catch!

src/components/pages/Account.tsx
markulrich: Yes, created https://github.com/rahafoundation/raha-mobile-app/issues/153

src/components/shared/VideoWithPlaceholder.tsx
markulrich: Just wanted to be consistent wherever we use `withNavigation`/`NavigationInjectedProps`. Initially I created an extra type everywhere PropsWithNavigation but seemed overly verbose

src/components/pages/Account.tsx
tinaroh: @osdiab  since I don't really know what normal style is!

PR #151, opened 2018-07-27
Enable send invite

tinaroh:
- Enables send invite
- Moves deep link param initialization to after LogIn since otherwise the user is not authorized to read the download URL
- Set invitingMember only if video download URL exists

### DIFF
--- a/src/components/pages/Onboarding/Onboarding.tsx
+++ b/src/components/pages/Onboarding/Onboarding.tsx
@@ -58,7 +58,6 @@ type OnboardingState = {
   verifiedName?: string;
   videoUri?: string;
   videoDownloadUrl?: string;
-  deeplinkInitialized?: boolean;
 };
 
 export class OnboardingView extends React.Component<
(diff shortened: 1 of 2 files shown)


src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -81,6 +81,10 @@ export class OnboardingView extends React.Component<
    * the associated video download url into state.
    */
   initializeDeeplinkingParams = async () => {
+    if (this.state.deeplinkInitialized) {
```
markulrich: I'm not clear on this, but is this a good place to use a setState callback since seems the goal is to call `setState` with `deeplinkInitialized: true` exactly once? https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

tinaroh: You're right, I moved it out of state!

PR #160, opened 2018-07-31
Add lastOpCreatedAt to calculate inactivity timer

markulrich:
Add lastOpCreatedAt to calculate inactivity timer and rename lastMinted to lastBasicIncomeAt

### DIFF
--- a/src/store/reducers/members.ts
+++ b/src/store/reducers/members.ts
@@ -70,7 +70,8 @@ interface RequiredMemberFields {
   fullName: string;
   createdAt: Date;
   inviteConfirmed: boolean;
-  lastMinted: Date;
+  lastMintedBasicIncomeAt: Date;
+  lastOpCreatedAt: Date;
 }
 
 type MemberFields = RequiredMemberFields & OptionalMemberFields;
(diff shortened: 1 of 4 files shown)


src/store/reducers/members.ts
```diff
@@ -305,14 +312,22 @@ function addMembersToState(
  * TODO (code cleanup): Avoid having to assert on id present for almost every operation type.
  */
 function applyOperation(
-  prevState: MembersState,
```
osdiab: I'd rather you just left this as `prevState` to not reassign parameters, and then later on just wrote

```
const newState = creator ?
  addMemberToState(prevState, creator.updateLastOpCreatedAt(createdAt)) : 
  prevState
```

and refer to that. no need to compromise here, gains nothing and muddies the meaning of the parameter (is it the original state or some working modified copy of it?).

src/store/selectors/members.ts
```diff
@@ -155,7 +158,3 @@ export function getMembersSortedByVotes(state: RahaState) {
     return voteCount === undefined ? 0 : voteCount;
   });
 }
-
-export function getDaysUntilInactive(member: Member) {
```
osdiab: was this not used anywhere?

PR #161, opened 2018-08-01
Only show people with invite confirmed in home feed

markulrich:


src/components/pages/Home.tsx
```diff
@@ -1,16 +1,57 @@
 import * as React from "react";
+import { Map } from "immutable";
+import { connect, MapStateToProps } from "react-redux";
 
 import { OperationType } from "@raha/api-shared/models/Operation";
-
+import { MemberId } from "@raha/api-shared/models/identifiers";
```
osdiab: I made a breaking change that I'll fix for you when the new `@raha/api-shared` version gets merged, but yeah this will just need to be `@raha/api-shared/dist/...` once #164 is in.

src/components/pages/Home.tsx
```diff
@@ -1,16 +1,58 @@
...
+          operation.op_code !== OperationType.MINT &&
+          isInviteConfirmed(props.membersById, operation.creator_uid)
+        }
+      />
+    </Container>
+  );
+};
 
-export const Home: React.StatelessComponent = () => {
+const UnconfirmedView: React.StatelessComponent<StateProps> = props => {
```
osdiab: I'm not totally clear on what this view is supposed to represent

PR #163, opened 2018-08-02
Fix iOS build

tinaroh:
Not sure what fixed it -- moving the mobile-Bridging-header didn't work so I removed react-native-video-processing and then re-added it (from the official GitHub account since my fix was merged) then went through all the steps again.

Seems to work now, please try with `yarn start:ios`. Seems to complain about Roboto missing but the app seems to work.


markulrich:
Yayay! Worked great on my emulator

PR #165, opened 2018-08-03
remove roboto font to fix ios build

osdiab:
not sure why roboto was ever imported to the app, but it doesn't exist in the fonts directory so ios build fails without this.


rahulgi:
Purge the Roboto!

PR #166, opened 2018-08-03
Add handling for new operation types to Redux state.

rahulgi:


src/store/reducers/members.ts
```diff
@@ -251,6 +312,62 @@ function applyOperation(
...
+        const newMember = new Member({
+          ...memberData
+        });
+
+        return addMemberToState(prevState, newMember);
+      }
+      case OperationType.REQUEST_VERIFICATION: {
+        const { to_uid } = operation.data;
+
+        assertMemberIdPresentInState(prevState, creator_uid, operation);
```
markulrich: Would it make sense to move all these `assertMemberIdPresentInState` into operationIsRelevantAndValid? Just at the beginning have something like if op in SET_CONST do the two asxertions?

src/store/reducers/members.ts
```diff
@@ -31,28 +34,45 @@ const GENESIS_TRUST_OPS = [
...
-function getDefaultMemberFields() {
+function getDefaultMemberFields(): OptionalMemberFields {
   return {
     balance: new Big(0),
     totalDonated: new Big(0),
     totalMinted: new Big(0),
     trustedBy: Set<MemberId>(),
     invited: Set<MemberId>(),
-    trusts: Set<MemberId>()
+    invitedBy: undefined,
```
osdiab: If you want to keep the lack of repetition, you can express this as  `undefined as MemberId | typeof GENESIS_MEMBER | undefined` and keep the old code

src/store/reducers/members.ts
osdiab: So long as the ids are predictable that would be fine

src/store/reducers/members.ts
rahulgi: TBH I like the explicit typedef. I find the above a little obtuse to parse lol

src/store/reducers/members.ts
rahulgi: I'll leave a TODO in for this for now - don't feel like modifying all the ops for this PR.

PR #170, opened 2018-08-07
Add the option to do joint vs async invite videos to Invite flow.

rahulgi:


src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -371,7 +378,7 @@ const mapStateToProps: MapStateToProps<ReduxStateProps, OwnProps, RahaState> = (
       state.authentication.isLoaded && state.authentication.isLoggedIn,
     deeplinkVideoToken,
     deeplinkInvitingMember,
-    isJointVideo: deeplinkInvitingMember ? true : false // TODO isJointVideo should be specified as part of the deeplink params
+    isJointVideo: deeplinkIsJointVideo ? true : false
```
markulrich: `!!deeplinkIsJointVideo` if need to force boolean

src/components/pages/Invite/InviteSplash.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface Props {
+  onContinue: () => void;
+  onBack: () => void;
+}
+
+export const InviteSplash: React.StatelessComponent<Props> = props => {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.back} onPress={props.onBack}>
```
tinaroh: Should this be a button? Also, why do we need an extra back button?

src/components/pages/Invite/Instructions.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isJointVideo: boolean;
+  onContinue: () => void;
+  onBack: () => void;
+}
+
+export const Instructions: React.StatelessComponent<Props> = props => {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.back} onPress={props.onBack}>
+        Back
```
tinaroh: Should this be a button? Also, why do we need an extra back button?

rahulgi: I just did a bit of text since I felt that a full button would be a bit too heavy. I removed the react-nav header, which is why I added our own back button - the react-nav back button wasn't properly backing out of the in-page flow and I felt that this looked nicer as a mostly full screen page.

src/components/pages/Onboarding/Onboarding.tsx
rahulgi: ah yes 👍

PR #176, opened 2018-08-09
Avoid keyboard on Give form

### LINKED ISSUE #135: Push up text box when filling out give form
Right now I can't see what I'm typing

tinaroh:
Fixes #135

The component `KeyboardAvoidingView` adjusts its padding when the keyboard is shown. But it doesn't work well with the navigators of `react-navigation` (the padding covers half the screen and doesn't actually push up the internal components of the screen) so I was unable to apply to the whole app.

I applied it to just the Give form so that when the user selects the bottom form input, it will scroll up so that the Give button is visible. The implementation is slightly hacky but I've tried different combinations for about 2 hours so I gave up on a cleaner solution for now...

`adjustPan` is so that the keyboard will cover the bottom tab navigator when the keyboard is up.

**Test Plan**
- Click on message form input on Give form, make sure Give button is still visible on both Android and iOS


src/components/pages/Give/GiveForm.tsx
```diff
@@ -153,102 +193,116 @@ class GiveFormView extends React.Component<Props, State> {
...
-            onPress={this.giveRaha}
-            disabled={
-              !this.validateForm() ||
-              (this.props.apiCallStatus &&
-                this.props.apiCallStatus.status === ApiCallStatusType.STARTED)
-            }
-          />
-        </View>
-      </Container>
+            <View style={{ height: 60 }} />
```
osdiab: What is this for?

tinaroh: oh forgot to remove it, thanks!

PR #177, opened 2018-08-09
Redirect via a loading splash page to handle deep-linked params.

rahulgi:


src/components/shared/Navigation.tsx
```diff
@@ -272,73 +275,100 @@ const ProfileTab = createTabNavigator(
...
+    initialRouteName: RouteName.LoadingSplash
   }
 );
 
-const SignedOutNavigator = createStackNavigator(
+const SignedOutNavigator = createSwitchNavigator(
   {
-    Onboarding: {
-      screen: Onboarding,
+    [RouteName.LoadingSplash]: {
```
tinaroh: Why is this an array?

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -212,15 +213,12 @@ export class OnboardingView extends React.Component<
...
-      return (
-        <React.Fragment>
-          <Text style={{ textAlign: "center" }}>
-            Welcome to Raha! Please sign up with your mobile number to accept
-            your invite.
-          </Text>
-          <LogIn navigation={this.props.navigation} />
-        </React.Fragment>
-      );
+      this.props.navigation.replace(RouteName.LogIn, {
```
tinaroh: Does back button still work with replace?

rahulgi: I used `replace` specifically to make sure that you can't hit the back button to go to onboarding, since that would leave you in the onboarding screen without having logged in.

src/components/shared/Navigation.tsx
rahulgi: Ah, it's kind of weird syntax but it's actually dereferencing LoadingSplash from RouteName and using that as the value of the key -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names

PR #185, opened 2018-08-11
Add ActivityItems for CreateMember, RequestVerification, and Verify ops

rahulgi:


src/components/shared/ActivityItem/CreateMemberOperationActivity.tsx
```diff
@@ -0,0 +1,61 @@
...
+  state,
+  ownProps
+) => {
+  const creatorId = ownProps.operation.creator_uid;
+  const createdMember = getMemberById(state, creatorId);
+
+  if (!createdMember) {
+    // TODO: log the following properly, properly handle cases when members are
+    // missing instead of throwing uncaught error
+    throw new Error(
```
osdiab: Created #187 to log this, it's already being done elsewhere so this is fine but we should probably do real logging and not crashingg.

PR #195, opened 2018-08-12
Activity Shape

osdiab:
Definte the data model of an Activity, and put all activity stuff into a folder.


src/components/shared/Activity/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * The content of an Activity. Missing some metadata that makes a complete,
+ * renderable Activity.
+ */
+export interface ActivityContent {
+  actor: Member;
+  description: (string | CurrencyValue)[];
+  body: ReactNode;
```
osdiab: Ah perhaps this ought to be agnostic of React, like everything else is. Basically the content body types are icon, set of videos/images, or post text based on my mockups.

osdiab: done

PR #200, opened 2018-08-18
Convert all operations to new activities format

osdiab:
Lots of restyling, and code to represent all operations as standardized Activity format.

Video: https://youtu.be/2M9UGhBvvZ8


src/store/selectors/activities/index.ts
```diff
@@ -0,0 +1,437 @@
...
+function activityContentContainsMember(
+  content: ActivityContent,
+  memberId: MemberId | typeof RAHA_BASIC_INCOME_MEMBER
+): boolean {
+  if (content.actor === RAHA_BASIC_INCOME_MEMBER) {
+    if (memberId === RAHA_BASIC_INCOME_MEMBER) {
+      return true;
+    }
+  } else {
+    if (content.actor.get("memberId") === memberId) {
```
markulrich: Use `else if`?

src/components/shared/Navigation.tsx
```diff
@@ -211,123 +248,127 @@ function createHeaderNavigationOptions(title: string) {
...
+    [RouteName.PendingInvitesPage]: PendingInvites
   },
   {
-    initialRouteName: RouteName.Profile
+    initialRouteName: RouteName.ProfilePage
   }
 );
 
+function getIconForRoute(routeName: RouteName): string {
+  switch (routeName) {
```
rahulgi: Maybe let's just make this a const map?

src/store/selectors/activities/types.ts
```diff
@@ -0,0 +1,128 @@
...
+  nextInChain?: {
+    direction: ActivityDirection;
+    content: ActivityContent;
+  };
+}
+
+/**
+ * Reference to another part of the app to redirect to.
+ * TODO: make params more specific.
+ * TODO: move this out of activities; colocate with RouteName
```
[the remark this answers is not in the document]
osdiab: Nah I think I should do it now so I will lol

src/store/selectors/activities/index.ts
```diff
@@ -0,0 +1,439 @@
...
+          }), invalid.`,
+          JSON.stringify(operation)
+        );
+        return memo;
+      }
+
+      switch (operation.op_code) {
+        case OperationType.CREATE_MEMBER: {
+          const newActivity: Activity = {
+            id: operation.id,
```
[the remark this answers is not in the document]
osdiab: They just need to be unique when they're rendered since they're used as keys for React, so maybe we can just make them a bunch of operation ids concatted. alternatively, I'll remove hte ID field, keep a list of participant operation ids along wiht the activity, and derive the key from that—probably better

src/store/selectors/activities/index.ts
```diff
@@ -0,0 +1,439 @@
...
+            timestamp: operation.created_at,
+            content: {
+              // type suggestions since GENESIS_MEMBER is only possible for
+              // TRUST operations
+              actor: creatorMember as Member,
+              description: ["just joined Raha!"],
```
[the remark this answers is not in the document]
osdiab: I don't see a value to these being constants, but they could be if you'd like

src/components/shared/Navigation.tsx
osdiab: done

src/store/selectors/activities/types.ts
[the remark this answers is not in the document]
osdiab: In fact this is an app-specific piece of functionality—if this data layer really is separate from the view layer it probably shouldn't refer to specific mobile route names, so that translation should be done at the view layer.

src/store/selectors/activities/index.ts
```diff
@@ -0,0 +1,456 @@
+import { Big } from "big.js";
+
+import {
```
rahulgi: Hmm, so this looks fine to me logic-wise. However, it does strike me that we're essentially running the same reduction that we perform on operations to generate the members state, including pretty much the same validation to check the validity of each operation before applying. And, this selector recalculates activities every time it gets called.

The trickiness seems to be that we're relying on members state to generate these activities. However, we only rely on members state as generated up-to the current operation we're creating the activity for.

Perhaps the "member" reducer could be come "memoized", and "applyOperation" could first apply the operation against memoized members state and then against memoized activity state.

src/components/shared/Activity/index.tsx
```diff
@@ -1,108 +1,366 @@
...
+            * flows correctly in descriptions.
+            */}
+          <Text style={styles.description}>
+            <MemberName member={actor} />
+            {description && <Text compoundContent={{ content: description }} />}
+          </Text>
+        </View>
+        {body && (
+          <View style={styles.contentBodyRow}>
+            {/* TODO: render chain direction */}
```
rahulgi: TODO satisfied.

src/components/shared/Activity/index.tsx
```diff
@@ -1,108 +1,366 @@
...
+            <View
```
rahulgi: Maybe this view could be abstracted out to a separate component?

src/components/shared/Activity/index.tsx
```diff
@@ -1,108 +1,366 @@
...
+    const { content, callToAction, timestamp } = this.props.activity;
+    return (
+      <View style={styles.activity}>
+        <ActivityContent
+          content={content}
+          onFindVideoElems={elems => {
+            this.videoElems = elems;
+          }}
+        />
+        {callToAction && <CallToAction callToAction={callToAction} />}
```
rahulgi: TODO: test CallToAction stuff

src/components/shared/Activity/index.tsx
```diff
@@ -1,108 +1,366 @@
...
+          <Text style={styles.timestamp}>
+            {formatRelative(timestamp, new Date()).toUpperCase()}
+          </Text>
+        </View>
+      </View>
+    );
+  }
 }
+
+const leftColumnWidth = 50;
```
rahulgi: When there's this much style info, how do you feel about a `index.styles.ts` file or something that we import from?

src/components/shared/Activity/index.tsx
```diff
@@ -1,108 +1,366 @@
...
+  return (
+    <View>
+      <MemberThumbnail style={styles.actorThumbnail} member={actor} />
+      {actions.map(action => (
+        <View>
+          {action.text.map(piece => {
+            if (typeof piece === "string") {
+              return <Text>piece</Text>;
+            }
+            if ("currencyType" in piece) {
```
rahulgi: I feel a bit weird about this mechanism of detecting what type something is and thus how it should be rendered. Maybe each of these action components should have an enum "type" associated with it that tells us how to render?

src/store/selectors/activities/index.ts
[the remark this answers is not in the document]
osdiab: I will add this as a TODO for later.

src/components/shared/elements/Text.tsx
```diff
@@ -7,19 +7,97 @@ import {
...
-import { fonts } from "../../../helpers/fonts";
+import { fonts, fontSizes } from "../../../helpers/fonts";
 import { colors } from "../../../helpers/colors";
+import { Currency, CurrencyValue } from "../Currency";
 
-export const Text: React.StatelessComponent<TextProps> = props => {
-  return <NativeText {...props} style={[styles.text, props.style]} />;
+export type CompoundContent = (string | CurrencyValue)[];
+
+function renderCompoundContent(
```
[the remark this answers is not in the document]
osdiab: done

src/components/pages/ReferralBonus/ReferralThumbnail.tsx
```diff
@@ -72,54 +79,41 @@ const ReferralThumbnailComponent: React.StatelessComponent<Props> = ({
   const actionButton = invitedMember.get("inviteConfirmed") ? (
     <Button
       onPress={mintReferralBonus}
-      loading={isMinting}
       disabled={isMinting}
-      title={`Mint +ℝ${REFERRAL_BONUS.toString()}`}
+      title={["Mint", REFERRAL_BONUS_VALUE]}
```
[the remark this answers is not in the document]
osdiab: I want to have control over how text gets styled in a button, which you can't do if it's arbitrary components (native doesn't cascade style rules like on the web). So I will compromise by forcing text to be a string.

src/store/selectors/activities/index.ts
osdiab: We can memoize selectors automatically with `reselect` so that it doens't have to be calculated every time, I'll save that for a future PR.

Secondarily the reduction is the same right now, because it's 1:1 between operations and activities, but very soon that will not be the case—it will be radically different once we start combining operations and stuff. The validity checks are much laxer here—just that once the member state has all been put together, that the resulting activites are actually valid given that state.

So I don't really agree with this comment, but I think that `reselect` would solve the performance concerns (as well as offloading this to the API, which is the ultimate endpoint).

src/components/shared/Activity/index.tsx
osdiab: That seems fine by me but not really that important

src/components/shared/Activity/index.tsx
osdiab: Good point, i'll save that for a future PR

osdiab: and leave a TODO

src/store/selectors/activities/index.ts
rahulgi: `reselect` sgtm.

-- Disclaimer -- both of the below concerns are probs not that big of a deal in light of us both doing serious validation of operations in the API itself and maintaining activities and member state in firebase instead of reconstructing it locally each time. That being said.. if you care to read the following:
---

Why should the validity checks be looser here? I don't see why we'd want to include operations that don't validate as we construct member state in constructing our activity state. In fact - I can construct a case right now where an op would validate against member state but not actually be valid -- any op that happens prior to the createMember operation, or, alternately, more sophisticated checks like "does the member have enough raha to perform this give operation?"

I also think that the reducer is being applied over the memoized state - so it's okay if operations and activities don't have a 1:1 relationship - the you can create 0, 1, many activities on the memoized state for each operation.

src/components/pages/Profile.tsx
```diff
@@ -51,143 +63,165 @@ type StatsProps = NavigationScreenProps<NavParams> & {
...
+          <Thumbnail member={member} />
+          <View style={styles.headerDetails}>
+            <Text style={styles.memberUsername}>@{member.get("username")}</Text>
+            <Stats navigation={navigation} member={member} />
+            {!!loggedInMember &&
+              !isOwnProfile && (
+                <View style={styles.memberActions}>
+                  <Button
+                    style={{ marginRight: 12 }}
+                    title="Trusted"
```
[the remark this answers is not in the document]
osdiab: fixed

src/components/pages/ReferralBonus/ReferralThumbnail.tsx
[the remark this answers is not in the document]
osdiab: made it use mixed content title value to force use of actual currency object.

src/store/selectors/activities/index.ts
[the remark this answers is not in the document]
osdiab: done

PR #210, opened 2018-08-21
Run yarn add appcenter and react-native link, set up $APPCENTER_SECRET

markulrich:


appcenter-pre-build.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
rahulgi: Does this script get called automatically during the build process?

osdiab:
IS this still experiencing the RNVideoProcessing issue? That usually occurs if you run `react-native link` without specifying the library you want to link, it causes react-native to try to re-link the `react-native-video-processing` library and then invariably does it wrong. If you did run that command, you can commit all changes to ios except for those that refer to `RNVideoProcessing` by running `git add -p`, and then check out the ios dir after committing to purge those erroneous `RNVideoProcessing` changes

appcenter-pre-build.sh
markulrich: Yep! Added comment

osdiab:
It failed because the dependencies weren't installed properly, so I installed them. Should be good now! running on my iPhone.

PR #220, opened 2018-08-23
Upgrade to RN Firebase 5.0 to support RN 0.56.

rahulgi:


tinaroh:
Hmm, I'm not getting any compilation issues on this branch. But the crash still happens. I'll investigate that separately. But this branch runs fine for me!

rahulgi:
Yuss, I think you fixed it, @tinaroh! Awesome!

PR #241, opened 2018-08-28
Fix function signature.

rahulgi:


osdiab:
I think `opFilter` is dangerous to have because once activities aren't 1:1 with operations this will almost certainly lead to inconsistent rendering of activities. But I'll look the other way for now. @markulrich  @rahulgi

PR #251, opened 2018-08-30
Fix video styling for verification videos.

rahulgi:
Not sure if there's a better way to do this, @osdiab. Kind of seems excessive to me to create a separate flex row just to center the video. Prior to this change the video was left-adjusted on the screen.


rahulgi:
screenshot_20180830-173847

PR #256, opened 2018-08-31
thumbnail generation, refactor verify flow

osdiab:
Generates thumbnails for all video uploads.

The same refactor probably ought to occur for other flows that don't use react-navigation, like onboarding and invite.

### DIFF
--- a/src/components/pages/Verify/index.tsx
+++ b/src/components/pages/Verify/index.tsx
@@ -2,17 +2,17 @@ import { BackHandler } from "react-native";
 import * as React from "react";
 import { NavigationScreenProps } from "react-navigation";
 import { MapStateToProps, connect } from "react-redux";
-import { RNFirebase } from "react-native-firebase";
 import DropdownAlert from "react-native-dropdownalert";
 
 import { MemberId } from "@raha/api-shared/dist/models/identifiers";
 
 import { IndependentPageContainer } from "../../shared/elements";
 import { VerifyCamera } from "./VerifyCamera";
-import { VideoPreview } from "../Camera/VideoPreview";
+import { VideoUploader } from "../../shared/VideoUploader";
 import {
   getLoggedInMember,
-  getAuthRestrictedVideoRef
+  getAuthRestrictedVideoRef,
+  getAuthRestrictedVideoThumbnailRef
 } from "../../../store/selectors/authentication";
 import { Member } from "../../../store/reducers/members";
 import { RahaState } from "../../../store";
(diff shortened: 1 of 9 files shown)


src/components/pages/Verify/index.tsx
```diff
@@ -41,56 +41,127 @@ type OwnProps = NavigationScreenProps<{ toMemberId: MemberId }>;
...
+ *    verifying the other person does not exist.
+ *
+ * 1. SPLASH: Informational page
+ * 2. CAMERA: Take a video verifying the other person.
+ * 3. CONFIRM_RECORDED_VIDEO: Confirm that the video is acceptable, or retake
+ *    (go back to CAMERA step). If it's good, then proceeding to the next
+ *    step uploads it.
+ * 4. CONFIRM_VERIFICATION: See above.
+ */
+type CurrentVerifyState =
```
rahulgi: nit: maybe name `VerifyStepState` and have the `VerifyStep` state members be `previousVerifyStepState` and `currentVerifyStepState`? I find that more clear than overloading `State`.

src/components/pages/Verify/index.tsx
```diff
@@ -182,52 +225,48 @@ class VerifyView extends React.Component<VerifyProps, VerifyState> {
...
             onRetakeClicked={() => {
-              this.setState({ step: VerifyStep.CAMERA });
+              this._handleNextStep({ step: VerifyStep.CAMERA });
             }}
             onError={(errorType: string, errorMessage: string) => {
               this.dropdown.alertWithType("error", errorType, errorMessage);
-              this.setState({
-                step: VerifyStep.CAMERA
-              });
+              this._handleNextStep({ step: VerifyStep.CAMERA });
```
rahulgi: Probs don't want to go anywhere on upload error.

tinaroh:
Hey Omar... I actually had some branches with thumbnail generation and unlike the previous thing that Rahul took from me, I got a bit further so we duplicated work.

Can we please follow the issue tracker? https://github.com/rahafoundation/raha-mobile-app/issues/208 I had the issue claimed, so could you please ping me first about it before starting work on it? Since you're done with this first and I don't think I'll be able to finish mine in the next two days this is totally fine but I could have told you my thoughts about it at least. It's for sure my fault that I've been really slow with work but it's pretty wasteful to be doing redundant work.

That aside, I don't think we should be passing two upload refs everywhere -- I think we should just pass getAuthRestrictedVideoBucketRef around and any video upload will generate the thumbnail. What do you think?

osdiab:
Oh, oops, yeah sorry about not checking the issues - got used to just talking to Rahul about what is going on but we should def make an effort to be more coordinated.

I can do that wrt to buckets, trying to make them separate was a reaction against thumbnail and video url locations being implicit - but I think that may only matter once they’re committed to operations. What I’d like ultimately is for all references to videos and thumbnails to explicitly pass locations, not implicitly have the names at well known locations; but since the endpoints don’t even take in any video and thumbnail locations, having them separate now isn’t so important.

tinaroh:
it's OK, not a big deal

PR #258, opened 2018-09-01
Update Discover (no more deep link confusion)

markulrich:


src/components/pages/Discover.tsx
```diff
@@ -82,6 +82,36 @@ function convertCardArr(cardArr: DiscoverCardRaw[]): DiscoverCard[] {
 
 // TODO below JSON should be available from website.
 const DISCOVER_INFO = convertCardArr([
+  {
+    header: "Check out Raha Services",
```
tinaroh: Why not marketplace?

markulrich: Yea maybe Marketplace is more clear ... trying to "nudge" people to offering Services but I agree wording is strange

PR #270, opened 2018-09-06
Launch Screen for iOS, rename generic react-native metadata

osdiab:
Replaces react-native default launch screen to one that looks like an empty page with the mint green header (iOS launch page is supposed to look like your app when it actually loads, to make the illusion of a fast loading app), also changes org name to Raha and renames `mobile` XCode app to Raha.

img_7649


tinaroh:
How about doing a splash page? All green with our logo in the center. A lot of apps do that

ios/Raha.xcodeproj/project.pbxproj
```diff
@@ -1116,7 +1116,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastUpgradeCheck = 610;
-				ORGANIZATIONNAME = Facebook;
```
rahulgi: lol

tinaroh: lollool

osdiab:
@tinaroh Apple explicitly says not to make it a splash page, but perhaps this is a case where nobody listens to what Apple actually wants you to do lololol. They say the purpose of a launch screen is to make it look like your app is loading faster, adn that it should only show up for fractions of a second on initial app launches and context switches; perhaps if we want to do that and our app load time is crappy then we should have a proper splash in app as well.

osdiab:
Would you prefer that I do that?

tinaroh:
I'd personally prefer a splash page over a blank page with just our header. But if the app loads really fast in iOS I don't care much. On Android it takes two seconds or so.

PR #272, opened 2018-09-06
don't allow anything but portrait until we good wtih landscape

osdiab:


rahulgi:
can we do this for android too?

osdiab:
done

PR #273, opened 2018-09-06
Remove code to handle deprecated REQUEST_INVITE operation.

rahulgi:


osdiab:
Is the web app still capable of hitting this? or has that been removed?

osdiab:
so long as ^^ is done then im good with this.

rahulgi:
it's been removed!

PR #286, opened 2018-09-07
Allow user to enter an invite token manually.

### LINKED ISSUE #261: Add a way for user to input invite token in-app in case deeplinking fails

tinaroh:
Closes #261.

Video showing switching between a joint invite token to async invite token then back:  https://goo.gl/MXvtsn

Error: 
screenshot_20180907-191050_raha test

### DIFF
--- a/src/components/pages/Onboarding/InputInviteToken.tsx
+++ b/src/components/pages/Onboarding/InputInviteToken.tsx
@@ -24,7 +24,7 @@ type OwnProps = {
 };
 
 type ReduxStateProps = {
-  invitationsState: InvitationsState;
+  doesInviteTokenExist: (t: string) => boolean;
 };
 
 type InputInviteTokenProps = OwnProps & ReduxStateProps;
@@ -40,7 +40,8 @@ const mapStateToProps: MapStateToProps<
   RahaState
 > = state => {
   return {
-    invitationsState: state.invitations
+    doesInviteTokenExist: (token: string) =>
+      state.invitations.byInviteToken.has(token)
   };
 };
 
@@ -56,7 +57,7 @@ class InputInviteTokenView extends React.Component<
       return;
     }
 
-    if (!this.props.invitationsState.byInviteToken.get(token)) {
+    if (!this.props.doesInviteTokenExist(token)) {
       this.setState({
         errorMessage: "This code is invalid or has expired."
       });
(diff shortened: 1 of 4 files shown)


src/components/pages/Onboarding/InputInviteToken.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * Page that requests an invite token
+ */
+
+type OwnProps = {
+  onInputInviteToken: (inviteToken: string) => any;
+  onBack: () => any;
+};
+
+type ReduxStateProps = {
+  invitationsState: InvitationsState;
```
osdiab: redux nit: I think this may be fine since it's just taking the type as a dependency and maybe everything in `InvitationsState` may really be relevant to this component, but in an ideal world components that subscribe to a redux store aren't supposed to depend on the redux store in any way besides the call to redux's `connect` function/stuff passed into it (like `mapStateToProps`, etc), so that they can be used/tested/understood independently of the redux store being available (for instance, if redux were no longer the tool we wanted to use, or if we wanted to use the same frontend code in some other context). in fact, the code they suggest puts that stuff into a separate file (called "containers") to be extra clear about it.

if this component is probably the only consumer of `InvitationsState` (and therefore extraneous state probably wouldn't get passed to it), we're not really planning on reusing this frontend code in other contexts, and this is just inheriting the shape of the state, then it's fine. 

if other stuff does end up getting into the state, then that may cause this component to experience unnecessary re-renders in the future because of changing state; at that case we'd want to enumerate directly what this is supposed to actually take, which would make the component more self-documenting as well.

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -68,35 +75,35 @@ type OnboardingState = {
...
-    if (this.props.hasValidInviteToken) {
-      this.initializeDeeplinkingState();
-    }
+    this.initializeDeeplinkingState();
   }
 
   /**
-   * If deeplinking params are present, makes sure they are valid, and fills in
-   * the associated video download url into state.
+   * Idempotently initializes deeplinking state. If deeplinking params are present,
```
osdiab: Sounds like it's not really initializing it, but something else (since initialization sounds like the kind of thing you do once). Might be a little confusing

src/components/pages/Onboarding/InputInviteToken.tsx
tinaroh: What would be a better way to use a variable from the component's state (the invitation token in the case) to check if it's in invitations? I only need to check that.

src/components/pages/Onboarding/Onboarding.tsx
tinaroh: It initializes once per token change though. Any suggestions on what to change it to?

PR #296, opened 2018-09-09
Make iOS Build Again

osdiab:


osdiab:
@markulrich check out these added lines of code very carefully

NOT lol there's 2.6mil of them. presumably from explicitly adding React deps to podfile and then installing, which code-push says we should do.

anyway FYI for it to work on your machine, you gotta a) in `ios/BuildConfigs/(build,release).xcconfig`, rename all references to `mobile` to `Raha`; then also run `yarn config:prod` or `yarn config:test`. Then it should run

Issue #276, opened 2018-09-07
Note current mint rate on mint page.


rahulgi:
A new user will probably be confused as to why they can't mint anything when they join.

markulrich:
Also amount able to mint increases by 0.01 about every 600 seconds, so could have a seconds countdown or pie chart if that would look nice and people wanna see something moving

PR #302, opened 2018-09-10
Add current mint rate to mint page.

rahulgi:
#276


rahulgi:
screenshot_20180910-173509

PR #304, opened 2018-09-10
Use keyboard avoiding scroll view instead of keyboardavoidingview

osdiab:
This is a more flexbile solution to keyboard avoidance imo.

~~It's a little awk in that you get scroll animations on the pages that use it~~ (bounce=false prevents this!), but it's pretty foolproof to issues of screen size, and easy to implement.

the one awkward UI thing is that when you scroll on the login page the image goes under the header, but that's pretty minor IMO and better for it to be usable than optimally pretty.

Videos:

* https://send.firefox.com/download/b6a51a1c36/#7CfcjID6xXI-MwND1ufCng
* https://send.firefox.com/download/377b8d985b/#Sp-0-lrLGNsplr8UruPtmg

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "react-native-dropdownalert": "^3.4.0",
     "react-native-elements": "^0.19.1",
     "react-native-firebase": "^5.0.0-rc1",
+    "react-native-keyboard-aware-scroll-view": "^0.7.2",
     "react-native-svg": "^6.5.2",
     "react-native-vector-icons": "^5.0.0",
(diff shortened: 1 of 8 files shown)


rahulgi:
haven't reviewed yet, but this firefox send ish is nice

rahulgi:
T.T looks like this on my phone

screenshot_20180910-212557

osdiab:
Lololol guess it’s not expanding to fill on android. Will look into it when
we’re together tomorrow :P

rahulgi:
tried debugging it, fml - #305

PR #308, opened 2018-09-12
Messaging infrastructure.

rahulgi:


src/backgroundMessaging.ts
```diff
@@ -0,0 +1,8 @@
+import firebase from "react-native-firebase";
+import { RemoteMessage } from "react-native-firebase/messaging";
+
+// Must return within 60 seconds.
+// See: https://rnfirebase.io/docs/v4.3.x/messaging/receiving-messages#1)-Add-Service-to-Android-Manifest
+export default async (message: RemoteMessage) => {
+  return Promise.resolve();
```
osdiab: Why are we not handling the message? looks like we're just throwing it away

osdiab: If there's a good reason for this, then please comment

src/components/MessagingManager.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+const mapStateToProps: MapStateToProps<
+  ReduxStateProps,
+  OwnProps,
+  RahaState
+> = state => {
+  const loggedInUid = getLoggedInFirebaseUserId(state);
+  if (loggedInUid) {
```
osdiab: can make it less nested with

```typescript
if (!loggedInUid) {
  return {};
}
// ... etc
```

src/store/actions/messaging.ts
```diff
@@ -0,0 +1,35 @@
...
+      const authToken = await getAuthToken(getState());
+      if (!authToken) {
+        throw new UnauthenticatedError();
+      }
+
+      callSetFcmToken(config.apiBase, authToken, fcmToken);
+    },
+    ApiEndpointName.SET_FCM_TOKEN,
+    // Arbitrary, we don't track these calls anywhere
+    "SET_FCM_TOKEN_CALL"
```
osdiab: Still make it an enum pl0x?

src/backgroundMessaging.ts
rahulgi: Removing these for now until we actually need them.

src/components/MessagingManager.tsx
```diff
@@ -0,0 +1,97 @@
...
+      if (this.props.loggedInMemberId) {
+        this.props.setFcmToken(fcmToken);
+      } else {
+        this.props.clearFcmToken(fcmToken);
+      }
+    }
+  };
+
+  async componentDidUpdate(prevProps: Props) {
+    if (prevProps.loggedInMemberId != this.props.loggedInMemberId) {
```
markulrich: Should we be using https://firebase.google.com/docs/reference/js/firebase.messaging.Messaging#onTokenRefresh to get new token whenever it changes?

src/store/reducers/apiCalls.ts
```diff
@@ -42,3 +42,14 @@ export const reducer: Reducer<ApiCallsState> = (
...
 };
+
+function getRandomInt(max: number) {
+  return Math.floor(Math.random() * Math.floor(max));
+}
+
+/**
+ * Returns a randomly generated identifier for an API call.
+ */
+export function generateRandomIdentifier() {
```
markulrich: Just to be doubly-sure we never use this instead of something like https://caniuse.com/#feat=getrandomvalues, let's add comment saying not safe for crypto.

src/components/MessagingManager.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  componentWillUnmount() {
+    if (this.unsubscribe) {
+      this.unsubscribe();
+    }
+  }
+
+  updateMemberFcmToken = async (newToken?: string) => {
+    const fcmToken = newToken || (await firebase.messaging().getToken());
```
markulrich: Should we cache `messaging` in https://github.com/rahafoundation/raha-mobile-app/blob/master/src/firebaseInit.ts#L22 to be consistent? (Or remove all of those if it's bad style?)

rahulgi: Sounds like a good idea to me!

src/components/MessagingManager.tsx
rahulgi: We are registering that callback above in `componentWillMount`.

PR #312, opened 2018-09-13
If user is auto logged in, show a dropdown alert that they were logged in

tinaroh:
https://drive.google.com/open?id=16O9TV3WbUJ6N3EJpbSwgIZZSycdpj8Co

The callback `verifyPhoneNumber` is more granular than `signInWithPhoneNumber` and allows us to detect when user is automatically verified.


tinaroh:
It's not a big deal for the video, it's meant to be public.

Issue #182, opened 2018-08-09
Clicking on a tab takes you to the last page on that tab's stack instead of the root page.


rahulgi:
Personally, I feel like it should go to the root page -- I don't want to see the profile of the last person I was looking at when I click the home tab, I want to see the feed.

markulrich:
With Instagram they do clicking takes you to the root IF you are already on that tab, so can double click to go to root. Feels like a good compromise and is probably what people are used to.

Right now we have that behavior on all tabs EXCEPT for Profile. I think we need to distinguish between generic Profile and ProfileRoot

PR #315, opened 2018-09-14
Give profile tab and profile page different route names

markulrich:
Makes it so that
- Back still works when you navigate to someone else's profile
- Hitting profile icon in bottom tab navigation correctly pops the stack


markulrich:
Fixes #182

PR #328, opened 2018-09-15
Install update on next resume

markulrich:


markulrich:
The default is only to install on restart which happens rarely. Other options include `IMMEDIATE` and `ON_NEXT_SUSPEND`

osdiab:
@rahulgi 's solution sounds reasonable to me. Should we add docs somewhere about it?

markulrich:
I agree `IMMEDIATE` seems like a bad experience and something we should avoid, which is why we shouldn't use mandatory unless there's a really good reason.

But for me `ON_NEXT_RESUME` seems like a good default? I don't anticipate often being in the middle of something, leaving the app, then quickly returning and being confused about things having changed

PR #329, opened 2018-09-16
Global dropdown alert component.

rahulgi:


src/components/DropdownWrapper.tsx
```diff
@@ -0,0 +1,95 @@
...
+import * as React from "react";
+import { MapStateToProps, connect } from "react-redux";
+import DropdownAlert from "react-native-dropdownalert";
+
+import { RahaState } from "../store";
+import { View } from "react-native";
+import { dismissDropdownMessage } from "../store/actions/dropdown";
+import { DropdownMessage } from "../store/reducers/dropdown";
+
+const AUTOMATICALLY_CLOSE_DROPDOWN_AFTER = 10000; // milliseconds
```
tinaroh: 10 seconds is a really long time... Should we do 3? What's the default?

src/store/reducers/dropdown.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface DropdownMessage {
+  type: DropdownType;
+  id: string; // autogenerated
+  title: string;
+  message: string;
+}
+
+export type DropdownState = {
+  messages: List<DropdownMessage>;
```
tinaroh: do we never clear this? should we remove message after we display it?

rahulgi: message gets cleared via the `DISMISS_DROPDOWN_MESSAGE` action which gets fired by the `DropdownWrapper` when the dropdown message is closed.

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -56,9 +57,17 @@ type ReduxStateProps = {
   hasValidInviteToken: boolean;
 };
 
+interface DispatchProps {
+  displayDropdownMessage: (
```
osdiab: Maybe we should just make this function signature a type in the dropdown actions so we don't have to repeat it everywhere, but whatevs

src/store/reducers/dropdown.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const reducer: Reducer<DropdownState> = (
+  prevState = {
+    messages: List()
+  },
+  untypedAction
+) => {
+  const action = untypedAction as DropdownAction;
+  switch (action.type) {
+    case DropdownActionsType.ADD_DROPDOWN_MESSAGE:
```
osdiab: Like I mentioned, we could alternatively do this by listening for other redux actions that directly correspond to stuff happening, but this approach is fine too. Logging for posterity

src/components/pages/Onboarding/Onboarding.tsx
osdiab: An alternative is to make `displayDropdownMessage` just take in a `DropdownMessage` argument rather than three sequential args (minus the autogenerated id, or with it optional at least) and then use that as the arg instead. I've been coming around to object args rather than sequential ones as being easier to work with since autocomplete happens and names are unambiguous, but either way works.

PR #330, opened 2018-09-16
Set up Branch custom domain link.raha.app

tinaroh:
We're using link.raha.app instead of d.raha.app to prevent downtime on existing apps. If we configure d.raha.app for Branch, we will not be serving up our previous webpage.

Omar, my iPad still doesn't work, can you verify https://link.raha.app/invite?t=abc1234 works for you when you get the chance?

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -61,6 +61,8 @@
             <data android:scheme="https" android:host="getraha-alternate.app.link" />
             <data android:scheme="https" android:host="getraha.test-app.link" />
             <data android:scheme="https" android:host="getraha-alternate.test-app.link" />
+            <data android:scheme="https" android:host="link.raha.app" />
+            <data android:scheme="https" android:host="link.test.raha.app" />
         </intent-filter>
       </activity>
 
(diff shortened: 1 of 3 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -61,6 +61,8 @@
             <data android:scheme="https" android:host="getraha-alternate.app.link" />
             <data android:scheme="https" android:host="getraha.test-app.link" />
             <data android:scheme="https" android:host="getraha-alternate.test-app.link" />
+            <data android:scheme="https" android:host="link.raha.app" />
```
markulrich: Is this what gets sent to user? Does branch require that we use `link` instead of something shorter?

tinaroh: Yes, it's what gets sent to the user. No, we can do something shorter. Suggestions that isn't "d"? We settled on "link" since it's pretty short, descriptive, and l is an ambiguous character

tinaroh: We're gonna do TO.raha.app!!! because it takes you TO the Raha App wowwww

PR #334, opened 2018-09-17
Respond to push notifications in the foreground.

rahulgi:


package.json
```diff
@@ -40,7 +40,7 @@
     "react-native-device-info": "^0.21.5",
     "react-native-dropdownalert": "^3.4.0",
     "react-native-elements": "^0.19.1",
-    "react-native-firebase": "^5.0.0-rc1",
+    "react-native-firebase": "^4.3.8",
```
tinaroh: Is this safe? I thought we needed 5.0.0-rc1 for RN 0.56

src/components/MessagingManager.tsx
```diff
@@ -26,31 +28,67 @@ interface ReduxStateProps {
...
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.loggedInMemberId != this.props.loggedInMemberId) {
+      this._updateMemberFcmToken();
+      this._setNotificationsListener();
     }
   }
 
-  updateMemberFcmToken = async (newToken?: string) => {
+  _check_permissions = async () => {
```
tinaroh: nit: camelcase

package.json
rahulgi: Yeah.. ideally we'd be using 5.0 :/ Maintainer says 4.3.8 is mostly okay though, a few edge cases where it's rough. https://github.com/invertase/react-native-firebase/issues/1481#issuecomment-421889666

tinaroh: The maintainer doesn't specify what the edge cases are though. Are you comfortable with downgrading? Or should we just wait 2 days?

tinaroh: Since this is just the foregrounded case I'm okay waiting.

rahulgi: down to wait two days i guess. we'll see how accurate the maintainer's time estimate is :P 

if the permissions request stuff turns out to be important for getting notifications on iOS to work, we can just pull them out and land earlier.

PR #335, opened 2018-09-17
iOS notification fixes.

rahulgi:
Omar: The problem was that the ios app registered in firebase had the wrong bundle id. This fixes that by using a newly registered ios app in our raha (and raha test) firebase projects.

~~This also includes a `pod install` for downgrading from v5 of firebase, if we're not downgrading then we should remove that commit.~~ @rahulgi  is removing the downgrade from this PR chain.

If iOS 12 also causes rn firebase to automatically use push notifications instead of captchas for human verification, which breaks because of this issue, then it would solve evan and hosea's problems too.

### DIFF
--- a/ios/Raha/Info.plist
+++ b/ios/Raha/Info.plist
@@ -46,7 +46,7 @@
 			<string>Editor</string>
 			<key>CFBundleURLSchemes</key>
 			<array>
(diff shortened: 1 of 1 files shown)


osdiab:
update: rahul is trying to remove the firebase downgrade from this PR, so i'll also remove the pod install

Issue #333, opened 2018-09-17
Automatically update state to include new operations


markulrich:

tinaroh:
Put out a PR to do the bare minimum.

How about doing a pull-to-refresh model instead or doing time intervals instead?

In the future, we should also do incremental updates for performance/

PR #336, opened 2018-09-17
Refresh ops every app foreground

tinaroh:
Makes #333 less urgent

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -17,19 +17,13 @@ const onBeforeLift = async () => {
   await refreshMembers()(store.dispatch, store.getState, undefined);
 };
 
-type AppRootProps = {};
-
-type AppRootState = {
+class AppRoot extends React.Component {
   shouldRefreshOnNextForeground: boolean;
-};
 
-class AppRoot extends React.Component<AppRootProps, AppRootState> {
-  constructor(props: AppRootProps) {
-    super(props);
-    this.state = {
-      // First refresh is from PersistGate#onBeforeLift
-      shouldRefreshOnNextForeground: false
-    };
+  constructor() {
+    super();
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -9,35 +9,75 @@ import { store, persistor } from "../store";
...
+  componentDidMount() {
+    AppState.addEventListener("change", this._handleAppStateChange);
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener("change", this._handleAppStateChange);
+  }
+
+  _handleAppStateChange = (nextAppState: string) => {
+    if (nextAppState === "background") {
```
rahulgi: Hmm, could we just compare the `nextAppState` value to `AppState.currentState` instead of maintaining the `shouldRefreshOnNextForeground` state ourselves?

i.e. in the handler, and using the documentation at https://facebook.github.io/react-native/docs/appstate.html:
```
const currentState = AppState.currentState;
if (
  // prevents us from refreshing on app start
  currentState &&
  nextAppState === "foreground" &&
  nextAppState !== currentState
) {
  refreshMembers()
}
```

tinaroh: hm, when I was testing, nextAppState === currentState...

rahulgi: I was reading https://medium.freecodecamp.org/where-do-i-belong-a-guide-to-saving-react-component-data-in-state-store-static-and-this-c49b335e2a00 and it says that component state that does not impact rendering should be stored on the class directly (i.e. `this.shouldRefreshOnNextForeground`), as state changes will trigger a re-render. I think we're doing similar things elsewhere in the app, I wonder what the performance implication of that is.

tinaroh: kk

PR #344, opened 2018-09-18
Version bump script, deployment instructions

### LINKED ISSUE #338: Update README_DEPLOYMENT with new deployment info
1. iOS is being published, deployment doc says it isn't
1. we're not going to use iOS build numbers, if you need to push a new build (i.e. because of app store rejection), increment the version number and build number is always 1
1. This doc is too long and anyone working with us including myself is too lazy to read all of it, in fact we should just put a TLDR deployment section in the main README and let this be an archive with more detailed snoozy information

### LINKED ISSUE #339: Create version bump script
That does the version logic described in #338 , so that we don't have to think much about pushing the right tags with the right versions properly

osdiab:
Script to bump version, and more thorough/less verbose deployment instructions.

Closes #338 
Closes #339

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,16 +77,20 @@
     "babel-preset-react-native": "^5.0.2",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "gradle-to-js": "^1.2.0",
     "jest": "^23.4.0",
     "jsdom": "^11.10.0",
+    "plist": "^3.0.1",
     "react-native-mock-render": "^0.0.25",
     "react-native-typescript-transformer": "^1.2.10",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -75,18 +76,24 @@
     "babel-jest": "^23.4.0",
     "babel-plugin-transform-remove-console": "^6.9.4",
     "babel-preset-react-native": "^5.0.2",
+    "chalk": "^2.4.1",
```
markulrich: Could all of these (and `typescript` for that matter) be a devDependency?

osdiab: They are `devDependencies`, it just looks like they're dependencies here because of the fold on Github.

PR #360, opened 2018-09-25
Make sure people with large text can find Join Now button

markulrich:


tinaroh:
Proposing #362 instead

PR #368, opened 2018-09-27
Add unverified label to MemberName component.

rahulgi:
screenshot_20180927-175042

### DIFF
--- a/src/components/pages/Profile.styles.ts
+++ b/src/components/pages/Profile.styles.ts
@@ -31,9 +31,14 @@ const statLabelStyle: TextStyle = {
   ...fontSizes.small
 };
 
-const memberUsernameStyle: TextStyle = {
+const memberNameStyle: TextStyle = {
   ...fonts.Lato.Semibold,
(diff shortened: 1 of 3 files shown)


rahulgi:
Not totally sure about the smaller font size.. what do you guys think?

rahulgi:
Also.. open to suggestions other than directly modifying the `MemberName` component.. though it seemed like the simplest way to make sure there aren't places where we forget to label a member.

tinaroh:
Hm I think it would be fine as the same font size actually since it looks kinda funky inline in the feed. I'm okay with sticking it in MemberName component

osdiab:
I also think that putting it next to the name is likely to run into issues with wrapping/running off the side, for people with really long names like you.

osdiab:
On further thought it doesn't seem scalable in general, so i added more ideas to that whimsical.

osdiab:
Talked offline, looks like Profile can use a rework altogether so for now this is a stopgap to change later.

PR #370, opened 2018-09-27
Add my SHA1s again for Firebase

tinaroh:
Sorry, in my overzealous attempt to clear disk space I nuked my Android debug.keystore so my fingerprint changed. Re-adding them here (don't remember which one my old one was...)

### DIFF
--- a/android/data/config/firebase/google-services.test.json
+++ b/android/data/config/firebase/google-services.test.json
@@ -39,6 +39,22 @@
           }
         },
         {
+          "client_id": "148482003030-1413e37njkqqkjdh85odmf3l9k44h6vk.apps.googleusercontent.com",
+          "client_type": 1,
+          "android_info": {
+            "package_name": "app.raha.mobileTest",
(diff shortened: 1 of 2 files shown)


android/data/config/firebase/google-services.test.json
```diff
@@ -60,10 +76,10 @@
               "client_type": 3
             },
             {
-              "client_id": "148482003030-485q62hqnch3m2mnue1ps0jdl35ikg7a.apps.googleusercontent.com",
+              "client_id": "148482003030-7anom7nbfkqlmk9oicnsmds7lsq3hbvc.apps.googleusercontent.com",
               "client_type": 2,
               "ios_info": {
-                "bundle_id": "app.raha.mobile"
+                "bundle_id": "app.raha.mobile.ios"
```
tinaroh: Not sure why this one changed :/

osdiab: that changed because `app.raha.mobile` for some reason was not ok with apple so i did that to make it unique when we first published

PR #371, opened 2018-09-27
Make onboarding pagination background black and add margins

tinaroh:
Sorry it probably looked better for me since my soft nav bar is gray.

Previous (off Rahul's phone):
screenshot_20180927-170021

Now:
screenshot_20180928-003012_raha test
screenshot_20180928-003103_raha test

### DIFF
--- a/src/components/shared/Swiper.tsx
+++ b/src/components/shared/Swiper.tsx
@@ -326,15 +326,15 @@ const styles = StyleSheet.create({
     flexGrow: 1
   },
   pagination: {
-    height: 12,
+    height: 20,
     width: "100%",
     flexDirection: "row",
     justifyContent: "center",
     alignItems: "center",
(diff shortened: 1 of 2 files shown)


rahulgi:
I think this is a small improvement, but would still like to see some of the changes that we talked about earlier. Approving though, since not that high pri.

tinaroh:
Sorry, which changes did you still want? There were a few ideas thrown out IIRC and none of them are ideal and we didn't agree on them.

I think we should just revamp the whole onboarding splash though since this page is getting skipped by most.

rahulgi:
Ah, I was thinking of same-color swiper and non-clashing "join now" button color.

tinaroh:
Yeah, I don't think that looks right since the swiper is not part of the button. I'll just keep this as-is for now.

PR #379, opened 2018-09-29
Enable editing of profile.

rahulgi:


src/store/selectors/activities/index.ts
```diff
@@ -108,6 +109,30 @@ function addCreateMemberOperationToActivites(
...
 
+function addEditMemberOperationToActivities(
+  state: RahaState,
+  activities: Activity[],
+  operation: EditMemberOperation
+): Activity[] {
+  const creatorMember = getOperationCreator(state, operation);
+  const description = operation.data.full_name
+    ? "edited their name."
+    : "edited their profile.";
```
tinaroh: What does "edited their profile" mean? Is it an error right now if that happens?

(Also, what sort of edits will EDIT_MEMBER operation support in the future? Does email and video fall under this?)

src/components/pages/AccountSettings/EditMember.tsx
```diff
@@ -0,0 +1,202 @@
...
+    if (!loggedInMember) {
+      return <Loading />;
+    }
+
+    const apiCallStatus =
+      editMemberApiCallStatus && editMemberApiCallStatus.status;
+
+    const disableGoBackButton = apiCallStatus === ApiCallStatusType.STARTED;
+    const goBackButtonTitle =
+      apiCallStatus === ApiCallStatusType.SUCCESS ? "Return" : "Cancel";
```
tinaroh: nit: Why "Return" instead of "Back"?

src/components/pages/AccountSettings/EditMember.tsx
```diff
@@ -0,0 +1,202 @@
...
+    return true;
+  }
+
+  onChangeFullName = (newName: string) => {
+    const { loggedInMember } = this.props;
+    const currentUsername = loggedInMember.get("username");
+    const currentFullName = loggedInMember.get("fullName");
+
+    if (newName === currentFullName) {
+      this.setState({
```
tinaroh: Does this need a special case? Is it for performance reasons?

src/store/selectors/activities/index.ts
rahulgi: Hmm.. I would say that right now it's unexpected, but not really an error. We could alternately just not create an activity if the operation doesn't have a new full name associated with it?

In the future, I don't think email will result in an EDIT_MEMBER operation, but changing the video could.

tinaroh: I think this is fine, was just wondering!

src/components/pages/AccountSettings/EditMember.tsx
rahulgi: Hmm.. I think of "Back" as more exiting out of a process and "Return" as what you do when you're done? Fine either way.

src/components/pages/AccountSettings/EditMember.tsx
rahulgi: `getUsername` generates a new username, while this case sets the username back to the current username if the full name is the same as the current full name.

rahulgi:
screenshot_20180930-023505
screenshot_20180930-023454

src/components/pages/AccountSettings/EditMember.tsx
markulrich: How about something like
```
this.setState({
        fullName: newName,
        username: newName === currentFullName ? currentUsername : getUsername(newName)
      })
```
?

rahulgi: lol what it's like you'd prefer we have nice code or something? derp ty lol XD

Issue #111, opened 2018-07-16
Add a build step to validate typescript.


rahulgi:

rahulgi:
Closing b/c CI now validates typescript.

PR #380, opened 2018-09-29
existing type errors and bug

osdiab:
There were a number of existing type errors in the code, a couple which caused bugs; this fixes those.

#378 also had some issues with how the types were defined—this rectifies that to prevent erroring out on displaying activities.

Could have been avoided by #111

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -17,11 +17,11 @@ const onBeforeLift = async () => {
   await refreshMembers()(store.dispatch, store.getState, undefined);
 };
 
-class AppRoot extends React.Component {
+class AppRoot extends React.Component<{}> {
   shouldRefreshOnNextForeground: boolean;
 
-  constructor() {
-    super();
+  constructor(props: {}) {
+    super(props);
     // First refresh is from PersistGate#onBeforeLift
     this.shouldRefreshOnNextForeground = false;
   }
--- a/src/components/shared/VideoUploader.tsx
+++ b/src/components/shared/VideoUploader.tsx
@@ -105,7 +105,7 @@ export class VideoUploader extends React.Component<
     // TODO figure out why local fetch is broken.
     // const response = await fetch(videoUri);
     // const blob = await response.blob();
-    //@ts-ignore Blob does not have data type
+    // (@)ts-ignore Blob does not have data type
     // if (blob.data.size > MAX_VIDEO_SIZE) {
     //   this.props.onError(
     //     "Error: Video Upload",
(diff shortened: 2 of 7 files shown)


src/components/App.tsx
```diff
@@ -17,11 +17,11 @@ const onBeforeLift = async () => {
...
 };
 
-class AppRoot extends React.Component {
+class AppRoot extends React.Component<{}> {
   shouldRefreshOnNextForeground: boolean;
 
-  constructor() {
-    super();
+  constructor(props: {}) {
+    super(props);
```
osdiab: IT wasn't happy that we weren't passing anything to `super`, because `React.Component` expects `props` as a param.

src/components/shared/VideoUploader.tsx
```diff
@@ -117,14 +117,14 @@ export class VideoUploader extends React.Component<
     // }
 
     // const metadata = {
-    //   //@ts-ignore Expo Blob does not have data type
+    //   // (@)ts-ignore Expo Blob does not have data type
     //   contentType: blob.data.type
     // };
 
-    const videoUploadTask = videoUploadRef.put(videoUri, {
+    const videoUploadTask = videoUploadRef.putFile(videoUri, {
```
osdiab: `putFile` is the proper method to call on `react-native-firebase` references. tested by uploading a verification video.

src/store/index.ts
```diff
@@ -6,31 +6,28 @@ import { Set as ImmutableSet, Map as ImmutableMap } from "immutable";
...
-  if (logged_props) {
+  if (!!logged_props) {
     const to_log = Object.keys(action)
-      .filter(key => logged_props.has(key))
-      .reduce((obj, key) => {
-        obj[key] = action[key];
-        return obj;
-      }, {});
-    const event = action.type.replace('.', '_');
+      .filter(key => logged_props && logged_props.has(key))
```
osdiab: This seems redundant, but throws a typescript error because the above check isn't triggering typescripts inference engine to narrow the type of `logged_props`. guess there's work for the typescript team to do there.

src/store/selectors/activities/types.ts
```diff
@@ -94,26 +99,22 @@ export type ActivityContent = {
    * display name in the format of a complete sentence.
    */
   description?: (string | CurrencyValue)[];
-} & (
```
osdiab: This doesn't express how our code actually works, since the `CREATE_MEMBER` operation actually does have a body, but no next in chain. But because of how typescript does union types, it seems it didn't actually throw that `body` was present but `nextInChain` wasn't in the activities selector; bummer.

src/store/selectors/activities/types.ts
```diff
@@ -94,26 +99,22 @@ export type ActivityContent = {
...
-    });
+  /**
+   * A larger, detailed body describing (in text) or representing (visually) the
+   * action.
+   *
+   * Currently, a body is required if there is to be further chained content. If
+   * that is too restrictive, we can relax that restriction, but it will likely
+   * require a visual redesign as well.
+   */
+  body?: {
```
osdiab: This causes a greater deal of nesting in the data structure, but more naturally represents that chained content comes after body content, always. if that's no longer the case at some point then we can split this up again, but the visual design wouldn't make that look particularly clean, so that would require some rethinking.

src/components/shared/VideoUploader.tsx
rahulgi: lol

PR #386, opened 2018-09-30
combine basic income mint activities

osdiab:
Currently it's set to combine all the mint activities in 3 hour windows starting from the timestamp of the first mint activity in that set.

Required changes to allow for multiple actors in a single activity, and caching so that we don't have to search the entire space of activities every time a new operation comes in—complexity ramping up.

Also filters out `request_verifications` operations in global feed to compensate for #387 not happening

## outstanding issues

None, but this will be reworked soon after according to [this comment below](https://github.com/rahafoundation/raha-mobile-app/pull/386#issuecomment-425796360).

## design questions

* @tinaroh  feels it is strange to total up multiple people's basic incomes; I think it makes sense from the perspective of giving scale to how much is being minted/what you're missing out on if you don't. what do you think?
* ~~can be saved for a future PR, but maybe we can add the time period of these mints? like "in the last 2 hours". adds immediacy~~ nahh doesn't really make sense—if a bunch of stuff happened like a week ago why would we say "they did it in the last 3 hours", not sensical.

### for later

* right now i just choose the first person's thumbnail to represent all the users, but maybe we can do a better representation like what fb messenger does, once we have actual thumbnails and not just letters.

### many people

img_7811

### two people

*(the line is for dramatic effect, but not actually in the app; i just derped when taking the screenshot lol)*

img_7812

### 1 person still works
img_7810

### DIFF
--- a/src/components/shared/Activity/index.tsx
+++ b/src/components/shared/Activity/index.tsx
@@ -248,9 +248,11 @@ class ActivityContent extends React.Component<{
             {actorsData === RAHA_BASIC_INCOME_MEMBER ? (
               <MemberName member={actorsData} />
             ) : (
-              (actorsData as Member[]).slice(0, 3).map((actor, index) => {
-                renderMemberNameInList({ actor, index, numActors });
-              })
+              (actorsData as Member[])
+                .slice(0, 3)
+                .map((actor, index) =>
+                  renderMemberNameInList({ actor, index, numActors })
+                )
             )}
             {/* TODO: make this clickable to see the list */}
             {numActors > 3 && (
--- a/src/components/shared/Activity/index.tsx
+++ b/src/components/shared/Activity/index.tsx
@@ -184,6 +184,31 @@ const ChainIndicator: React.StatelessComponent<{
   );
 };
 
+function renderMemberNameInList({
+  actor,
+  index,
+  numActors
+}: {
+  actor: Member;
+  index: number;
+  numActors: number;
+}): React.ReactNode {
+  const insertComma = numActors > 2 && index < Math.min(numActors, 4) - 2;
+  const insertAnd = numActors > 1 && index === Math.min(numActors, 4) - 2;
+  return (
+    <React.Fragment key={index}>
+      <MemberName member={actor} />
+      {insertComma && <Text>, </Text>}
+      {insertAnd && (
+        <Text>
+          {!insertComma && " "}
+          and{" "}
+        </Text>
+      )}
+    </React.Fragment>
+  );
+}
+
 class ActivityContent extends React.Component<{
   content: ActivityContentData;
   onFindVideoElems: (elems: VideoWithPlaceholderView[]) => void;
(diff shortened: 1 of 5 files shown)


tinaroh:
Maybe we can omit grouping new activities for some time frame? Like only recombining after app launches or app starts.

src/store/selectors/activities/index.ts
```diff
@@ -277,13 +359,58 @@ function addMintOperationToActivities(
...
   switch (operation.data.type) {
     case MintType.BASIC_INCOME: {
+      // if cache is present, consider combining this basic income mint
+      // operation with existing activity
+      if (combineActivitiesCache.aggregateBasicIncome) {
+        // check if difference in time is too long before merging
+        const existingActivityId =
+          combineActivitiesCache.aggregateBasicIncome.aggregatedActivityId;
+        const existingActivity = activities.get(existingActivityId);
+        if (!existingActivity) {
```
osdiab: Rather than doing this check I guess I can just store the full activity in the cache.

src/components/shared/Activity/index.tsx
```diff
@@ -186,22 +191,59 @@ class ActivityContent extends React.Component<{
   ownVideoElems: VideoWithPlaceholderView[] = [];
 
   public render() {
-    const { actor, description, body } = this.props.content;
+    const { actors, description, body } = this.props.content;
+    const actorsArray: [typeof RAHA_BASIC_INCOME_MEMBER] | Member[] =
```
rahulgi: This is saying it could either be one `RAHA_BASIC_INCOME_MEMBER` or an array of real members? Is there a cleaner way to do that?

src/components/shared/Activity/index.tsx
```diff
@@ -186,22 +191,59 @@ class ActivityContent extends React.Component<{
...
+            member={actorsArray[0]}
           />
           {/*
             * Everything in the description must ultimately be Text elements, or
             * else React Native doesn't support it. This is done to ensure text
             * flows correctly in descriptions.
             */}
           <Text style={styles.description}>
-            <MemberName member={actor} />
+            {/*
```
rahulgi: Eh, can we move this into a helper, please? :)

src/store/selectors/activities/index.ts
```diff
@@ -259,15 +279,76 @@ function addGiveOperationToActivities(
...
+    timestamp:
+      existingActivity.timestamp > operation.created_at
+        ? existingActivity.timestamp
+        : operation.created_at,
+    content: {
+      ...existingActivity.content,
+      description: ["minted a total of", totalMinted, "of basic income."],
+      // TODO: show the most relevant members to the logged in member first, not
+      // just in the order they're found
+      // Only add the actor if they aren't already in the list, to prevent
```
rahulgi: We're not actually performing this check, right? It happens automatically b/c we're using a map? I find this comment a bit confusing.

osdiab:
This will be rewritten to simplify it by splitting apart the concept of related operations from the concept of displayed stories, [as discussed in Slack](https://rahainstitute.slack.com/archives/CBSDMAFPX/p1538371156000100). But in the interest of having this functionality for demoing purposes, we'll merge this.

osdiab:
oh it looks like @markulrich  and @tinaroh don't want to show the total amount, should I just remove that for now?

osdiab:
Squelched it for now.

PR #387, opened 2018-09-30
remove request_verification activities

osdiab:
according to this beautiful flowchart I made (which we should probably post somewhere for posterity), request_verification activities don't need to be displayed—and right now they're producing duplicate videos, so might as well remove now,

screen shot 2018-10-01 at 12 42 54 am

### DIFF
--- a/src/store/selectors/activities/index.ts
+++ b/src/store/selectors/activities/index.ts
@@ -135,47 +135,6 @@ function addEditMemberOperationToActivities(
   return [...activities, newActivity];
 }
 
-function addRequestVerificationOperationToActivites(
-  state: RahaState,
-  activities: Activity[],
-  operation: RequestVerificationOperation
-): Activity[] {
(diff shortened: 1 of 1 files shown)


osdiab:
@rahulgi a little confused what `request_verification` is for now that i'm thikning about it a bit

so `invite` doesn't contain the `invite_token`? and if `create_member` already contains the inviting member, why can't it contain whatever is in `request_verification`?

rahulgi:
I broke request_verification out into a separate operation since I thought that we might want to allow people to explicitly request that other people / verifiers verify their accounts.

That being said.. I didn't intend that request_verification be something that gets flushed to the blockchain, but right now, since we construct state from operations, the operation stream is our hammer. I do think it's important that "request_verification" show up in the requested user's feed, at least, since there is no other persistent way for them to know that they need to go and verify your profile until we have a separate notification stream.

osdiab:
Closing due to reason above.

PR #394, opened 2018-10-02
Fix feed crash

tinaroh:
EDIT_MEMBER op was not compatible with the bundling changes

It wouldn't have generated a merge conflict. Maybe we need CI to build, make sure it builds before merge?


osdiab:
Ya we should probably set up CI would save us from a bunch of stuff, also should probably check if code passes with merge or require a merge from master before merging in?

PR #396, opened 2018-10-03
Allow users to give amounts with decimal points

tinaroh:
By rounding with Big after each character, we remove all decimal points that the user attempts to enter;

Instead, I now validate and round after onEndEditing is called when the user selects away from the amount FormInput.

Demo: https://drive.google.com/file/d/1lQFRbco-sloEyUsf-YxaWUMtvOq_OPNb/view?usp=sharing

### DIFF
--- a/src/components/pages/Give/GiveForm.tsx
+++ b/src/components/pages/Give/GiveForm.tsx
@@ -201,7 +201,7 @@ class GiveFormView extends React.Component<Props, State> {
           <FormLabel labelStyle={styles.formLabelStyle}>Amount</FormLabel>
           <FormInput
             inputStyle={styles.formInputAmountStyle}
-            keyboardType="numeric"
+            keyboardType="decimal-pad"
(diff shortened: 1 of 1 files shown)


osdiab:
@rahulgi @markulrich I think it makes sense to round when displaying stuff for ease of use but is there a reason we disallow people from using more decimal points? I'm fine with this change since we are rounding things in the backend. but want to know why we decided to actually force rounding like that

tinaroh:
I think letting the user know how much they're actually going to be spending is helpful, especially if we round up.

osdiab:
I agree, i'm just asking why we're rounding up in the actual backend / hard not allowing fractional sends. UX-wise in the app makes sense like how Venmo rounds, but backend seems kinda arbitrary since for instance you can transact in fractions of pennies, and for a lot of cryptos it's essential for trading reasonable values like btc

PR #413, opened 2018-10-08
Refactor Activities into Stories

osdiab:
This PR does quite a bit.

1. `Activities` now refer to just conceptually whole related operations. They contain no information about how to render them, they mostly are just sets of related operations.
1. `Stories` act as a data format for how to present `Activities`, and may be comprised of one or more `Activities` (like mint basic income activities that get displayed together). This could have been done entirely within components but I like the flexibility of working with plain Javascript objects rather than having complex/not portable business logic in React components.
1. Combining `CREATE_MEMBER`, `VERIFY_MEMBER` and `MINT_REFERRAL_BONUS` operations has been completed.
1. However, `VERIFY_MEMBER` videos are being hidden for now until a) we have the ability to detect duplicates, primarily in the case of joint invite videos; and b) until we have side by side display of media, so that they don't take up the entire screen when they're presented.
1. With this logic, all the messed up ordering of timestamps should be fully resolved.

## Screenshots

img_7855
img_7854
img_7852


src/components/shared/StoryFeed/index.tsx
```diff
@@ -6,30 +6,32 @@
 import * as React from "react";
 import { FlatList, FlatListProps } from "react-native";
 
-import { Activity, ActivityView } from "./";
+import { Story, StoryView } from "../Story";
 import { Activity as ActivityData } from "../../../store/selectors/activities/types";
+import { Story as StoryModel } from "../../../store/selectors/stories/types";
+import { List } from "immutable";
 
 interface ActivityFeedProps {
```
tinaroh: nit: StoryFeedProps

src/store/selectors/stories/types.ts
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Reference to find and render an image.
+ */
+export interface ImageReference {
+  imageUri: string;
+}
+
+/**
+ * Reference to visual media to dipslay.
```
tinaroh: super nit sp: display

src/store/selectors/stories/types.ts
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * An invitation for a member to take action.
+ */
+export type CallToAction = CallToActionPiece[];
+
+/**
+ * Shape of the possible data that can comprise a story
+ */
+export interface StoryDataDefinition<
```
tinaroh: hmm why are `StoryContent` and `StoryData` separated? If `StoryData` is  always Activity/Activities, can we call it `StoryActivity` instead?

osdiab: the `type` sets it apart from just a list of activities, since some activities can be intepreted in different ways—a mint operation is treated differently if it's a referral or a mint, a verify operation is treated differently if it's alone vs tied to a new member scenario. I wanted `StoryData` to unambiguously define how the actual visual content—the `StoryContent`—gets generated.

src/components/shared/Navigation.tsx
```diff
@@ -202,12 +202,12 @@ const MemberList = {
...
 
-const ActivityList = {
-  screen: ActivityListScreen,
+const StoryList = {
+  screen: StoryListScreen,
   navigationOptions: ({ navigation }: any) => {
     return {
       headerTitle: (
-        <HeaderTitle title={navigation.getParam("title", "Activity List")} />
+        <HeaderTitle title={navigation.getParam("title", "Story List")} />
```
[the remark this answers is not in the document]
osdiab: yeah i think it's a default if no title was passed. I'll change it to Stories. I'm not worried about people being confused about it vs Insta/Snap stories, even those features are quite different and I don't think anyone's confused that Raha is not those.

src/store/selectors/stories/types.ts
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Represents the direction of the relationship between actors in an story.
+ * Currently only supports onward or bidirectional, since that makes the flow
+ * of the application linear and therefore more comprehensible.
+ */
+export enum ChainDirection {
+  Forward = "Forward",
+  // Backward = "Backward",  // I think Backwards messes with the flow of the feed.
+  Bidirectional = "Bidirectional",
```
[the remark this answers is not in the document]
osdiab: Bidirectional is a real word lol. No need for weird casing.

src/store/selectors/stories/types.ts
osdiab: added more docs to explain this.

tinaroh: Thanks for the updated docs!

src/components/shared/Navigation.tsx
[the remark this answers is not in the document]
tinaroh: *cue me being confused* T_T

PR #414, opened 2018-10-09
Larger mint buttons, fix Mint Invite Bonuses header

markulrich:
Omar: Updated to make it look better.

### Screenshots

img_7862

img_7863

### DIFF
--- a/src/components/pages/Wallet.tsx
+++ b/src/components/pages/Wallet.tsx
@@ -1,5 +1,12 @@
 import * as React from "react";
-import { StyleSheet, View, Image, TextStyle, ViewStyle } from "react-native";
+import {
+  StyleSheet,
+  View,
+  Image,
+  TextStyle,
+  ViewStyle,
+  ScrollView
+} from "react-native";
 import { connect, MapStateToProps } from "react-redux";
 
 import { MemberId } from "@raha/api-shared/dist/models/identifiers";
@@ -177,10 +184,10 @@ const Invite: React.StatelessComponent<Props> = props => {
 
 const WalletView: React.StatelessComponent<Props> = props => {
   return (
-    <View style={styles.page}>
+    <ScrollView bounces={false} contentContainerStyle={styles.page}>
       <MoneySection {...props} />
       <Actions {...props} />
-    </View>
+    </ScrollView>
   );
 };
 
@@ -198,17 +205,17 @@ const numberLabelStyle: TextStyle = {
 };
 
 const inviteSectionTextStyle: TextStyle = {
-  marginVertical: 24,
+  marginVertical: 18,
   textAlign: "center"
 };
 
 const buttonStyle: ViewStyle = {
-  marginVertical: 8
+  marginVertical: 12
 };
 
 // shared to create consistent spacing
 const sectionSpacer: ViewStyle = {
-  marginTop: 20
+  marginTop: 15
 };
 
 // pushes content on mint page down so image shows up on bottom rather than
@@ -220,7 +227,7 @@ const pushDownSpacerStyle: ViewStyle = {
 const financesSectionStyle: ViewStyle = {
   flex: 0, // don't expand to fill space
   alignSelf: "stretch", // take full width
-  marginBottom: 20,
+  marginBottom: 10,
 
   // balance on left side, other financial info pushed to right.
   flexDirection: "row",
@@ -236,11 +243,10 @@ const mintButtonStyle: ViewStyle = { ...sectionSpacer, ...buttonStyle };
 
 const pageStyle: ViewStyle = {
   backgroundColor: colors.pageBackground,
-  flex: 1,
+  minHeight: "100%",
   padding: 20,
   flexDirection: "column",
-  alignItems: "center",
-  justifyContent: "space-between"
+  alignItems: "center"
 };
 
 const actionsSectionStyle: ViewStyle = {
@@ -257,9 +263,10 @@ const actionsSectionGetVerifiedTextBlockStyle: TextStyle = {
 
 const actionImageStyle: ViewStyle = {
   marginTop: 8,
-  // shrink images to ensure screen doesn't overflow
-  flex: -1,
+  flex: -1, // allow the image to shrink up to the min height
   flexBasis: 200,
+  maxHeight: 200,
+  minHeight: 150,
   maxWidth: "100%"
 };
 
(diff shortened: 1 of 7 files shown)


markulrich:
simulator screen shot - iphone 6 - 2018-10-09 at 00 23 16

simulator screen shot - iphone 6 - 2018-10-09 at 00 23 04

src/components/shared/MintButton.tsx
```diff
@@ -74,15 +75,21 @@ const MintButtonComponent: React.StatelessComponent<Props> = props => {
...
               role: CurrencyRole.None
             },
-            "per week."
+            "a week, limit",
+            {
+              currencyType: CurrencyType.Raha,
+              value: new Big(RAHA_MINT_MAX),
+              role: CurrencyRole.None
+            },
+            "per MINT."
```
osdiab: all caps mint is weird. let's make it

"You may create/mint R10.00 per week up to a maximum of R40.00 at once."

Or maybe,

Basic Income Rate: R10.00 per week
Max outstanding basic income: R40.00

src/components/shared/MintButton.tsx
```diff
@@ -74,15 +75,21 @@ const MintButtonComponent: React.StatelessComponent<Props> = props => {
       />
       <Text style={{ marginTop: 4 }}>
         <MixedText
-          style={[fontSizes.small]}
+          style={[fontSizes.small, { textAlign: "center" }]}
           content={[
-            "The current mint rate is",
+            "Create ",
```
osdiab: Do we want to use the word "Create" or "Mint"? We keep switching between the two, i think we should choose one and commit. Mint is more conceptually compelling but Create is clearer.

src/components/pages/Wallet.tsx
```diff
@@ -140,16 +141,14 @@ const Invite: React.StatelessComponent<Props> = props => {
...
   return (
     <React.Fragment>
-      <Text style={[styles.inviteSectionText]}>
-        You have nothing to mint at this time.
-      </Text>
+      <Text style={[styles.inviteSectionText]}>Nothing to mint right now.</Text>
       <View>
         <Text style={styles.inviteSectionText}>
-          Invite a friend to earn 60 Raha:
+          Invite a friend to create {REFERRAL_BONUS} Raha!
```
osdiab: Same mention here, inconsistent wording: Mint or Create?

src/components/pages/Wallet.tsx
```diff
@@ -140,16 +141,14 @@ const Invite: React.StatelessComponent<Props> = props => {
...
         </Text>
         <Button
           style={styles.button}
-          title="Invite"
+          title={`INVITE +${REFERRAL_BONUS}`}
```
osdiab: You don't need to capitalize it, use normal capitalization: Invite +${REFERRAL_BONUS}

src/components/pages/Wallet.tsx
```diff
@@ -187,7 +186,9 @@ const inviteSectionTextStyle: TextStyle = {
 };
 
 const buttonStyle: ViewStyle = {
-  marginVertical: 8
+  marginVertical: 8,
+  paddingVertical: 16,
+  width: 300
```
tinaroh: minWidth? This will probably break depending on font size / i18n

osdiab: why does it need an explicit width? can't it just be made consistently good with proper padding? What motivated this change?

osdiab: if you want it to be full width you can probably just go with `width: "100%"` or something using `flex` sizing to grow to full width, depending on how this page's styles are structured. One-off random numbers like this is how we end up with giant styling headaches.

src/components/pages/Wallet.tsx
```diff
@@ -140,18 +141,16 @@ const Invite: React.StatelessComponent<Props> = props => {
...
+      <Text style={[styles.inviteSectionText]}>Nothing to mint right now.</Text>
       <View>
         <Text style={styles.inviteSectionText}>
-          Invite a friend to earn 60 Raha:
+          Invite a friend to mint {REFERRAL_BONUS} Raha!
         </Text>
         <Button
           style={styles.button}
-          title="Invite"
+          title={`Invite +${REFERRAL_BONUS}`}
```
osdiab: We can actually make it have the full R 60.00, like how `MintButton` does it; you can do something like `title={["Invite +", referralBonus]}` where `referralBonus` is the actual Currency (with `CurrencyRole.None` to use the same color) object referring to that referral bonus. That makes it clearer it's an amount of currency and is also more consistent

src/components/pages/Wallet.tsx
```diff
@@ -187,7 +186,9 @@ const inviteSectionTextStyle: TextStyle = {
 };
 
 const buttonStyle: ViewStyle = {
-  marginVertical: 8
+  marginVertical: 8,
+  paddingVertical: 16,
+  minWidth: 300
```
osdiab: You can push with this, but why still `300`? If you want it to be full width I think there are better ways than this... I think `width: 100%` should work, but if it doesn't you can put the button in a `View` with `flexDirection: row` and giving the button `flex: 1`. The parent view would automatically fill to full width and the button would then grow to full width of the parent. `300` might look good on your device but isn't flexible to others and makes for inconsistent visuals in the app.

osdiab: If you just don't want to deal with this right now, then I'll make the button component more expressive to our needs later on. The fact you'd like to change the padding for this button and this one only seems to reflect that the button styles are not generalizing to different situations, so having choices for standardized button sizing is probably useful as well.

osdiab: although from the screenshots i cant actually tell what changing the button padding is doing. can you just remove that line?

markulrich:
simulator screen shot - iphone 6 - 2018-10-10 at 00 06 38

src/components/shared/Navigation.tsx
```diff
@@ -372,7 +372,7 @@ const WalletTab = createNavigatorForTab(
     },
     [RouteName.ReferralBonusPage]: {
       screen: ReferralBonus,
-      navigationOptions: createHeaderNavigationOptions()
+      navigationOptions: createHeaderNavigationOptions("Mint referral bonus", true)
```
tinaroh: Can we just call it Referrals? Seems like a long header that could get chopped off on smaller phones

osdiab: done

src/components/pages/Wallet.tsx
osdiab: fixed this.

osdiab:
@tinaroh please take a look, i made changes, see what it looks like from screenshots on the description of the issue.

PR #421, opened 2018-10-10
Improve logging

markulrich:


src/components/pages/LogIn.tsx
```diff
@@ -403,12 +403,12 @@ class LogInView extends React.Component<LogInProps, LogInState> {
     }
 
     // set the sent time for the resend timer, which occurs when you get to
-    // WAITING_FOR_CONFIRMATION_INPUT state (see the initiatePhoneLogIn action)
+    // WAITING_FOR_CONF_INPUT state (see the initiatePhoneLogIn action)
```
tinaroh: I actually prefer more verbose with no disambiguity.. was there a character limit?

src/components/pages/LogIn.tsx
```diff
@@ -455,24 +455,24 @@ class LogInView extends React.Component<LogInProps, LogInState> {
...
-      case PhoneLogInStatus.SENDING_PHONE_NUMBER_FAILED:
+      case PhoneLogInStatus.WAITING_FOR_PHONE_INPUT:
+      case PhoneLogInStatus.SENDING_PHONE:
+      case PhoneLogInStatus.SENDING_PHONE_FAILED:
         return (
           <PhoneNumberForm
             waitingForCode={
               this.props.phoneLogInStatus.status ===
-              PhoneLogInStatus.SENDING_PHONE_NUMBER
+              PhoneLogInStatus.SENDING_PHONE
```
tinaroh: Sending phone doesn't make much sense to me

src/store/actions/authentication.ts
```diff
@@ -9,20 +9,21 @@ import { auth } from "../../firebaseInit";
 import { config } from "../../data/config";
 import branch from "react-native-branch";
 
+// Keep these names to 40 characters or less so they may be used for Firebase Analytics events
```
markulrich: Cc @tinaroh here's the ones I had to change

src/components/pages/LogIn.tsx
markulrich: There is https://firebase.google.com/docs/reference/cpp/group/parameter-names 40 char limit, but I'll go ahead and make ones where we can and undo changes to `PhoneLogInStatus` since we aren't logging that

src/components/pages/LogIn.tsx
markulrich: will change back

PR #432, opened 2018-10-12
Add subscribe to monthly newsletter in onboarding flow

tinaroh:
screenshot_20181012-145741_raha test


src/components/pages/Onboarding/OnboardingCreateAccount.tsx
```diff
@@ -42,6 +43,7 @@ type OwnProps = {
   emailAddress: string;
   videoToken: string;
   inviteToken?: string;
+  subscribeToNewsletter?: boolean;
```
rahulgi: should this be required?

PR #441, opened 2018-10-14
Add hint icon component

tinaroh:
It shows a gray icon with "i" (default) or "?" and when you click it, it pops up a modal. Will use this to provide more info to the user throughout the app.

### DIFF
--- a/src/components/shared/elements/Hint.tsx
+++ b/src/components/shared/elements/Hint.tsx
@@ -5,7 +5,8 @@ import {
   View,
   ScrollView,
   StyleSheet,
-  TextStyle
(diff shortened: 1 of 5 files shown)


src/components/shared/elements/Hint.tsx
```diff
@@ -0,0 +1,88 @@
...
+            size={20}
+            color={colors.modalIcon}
+          />
+        </TouchableOpacity>
+      </React.Fragment>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
```
rahulgi: nit: explicitly typed styles

PR #465, opened 2018-10-23
Fix app name for Android

tinaroh:


rahulgi:
@tina this is changing the published bundle ID? Will the app store transparently handle this change for people who already have the app installed?

tinaroh:
We were already publishing with `assembleProdRelease`. We're not changing anything. But for some reason the productFlavor settings were not getting applied when I built without this change so when you open activity manager the name of the app was the default `app.raha.mobile.MainActivity` or whatever. 

I can try to dig more into why it's not getting applied though.

tinaroh:
BTW, bundle ID is an iOS concept. Product flavor is not something shown to the end user, it's a way to tell the build system what configurations to use.

rahulgi:
ah okay, gotcha. LGTM, thanks!

tinaroh:
I'm actually going to close this. Just built again and wasn't able to reproduce my issue from yesterday... Maybe something up with my phone...

PR #485, opened 2018-11-02
Full page invite token page during onboarding.

rahulgi:
Also forces all new members to create an identity video, regardless of whether they were invited via joint video.
Also fixes a bug in redirecting during onboarding.

screen

### DIFF
--- a/src/components/pages/LogIn.tsx
+++ b/src/components/pages/LogIn.tsx
@@ -446,7 +446,7 @@ class LogInView extends React.Component<LogInProps, LogInState> {
       const redirectParams = this.props.navigation.getParam("redirectParams");
       this.props.navigation.navigate(
         redirectTo,
-        ...(redirectParams ? redirectParams : {})
+        redirectParams ? redirectParams : {}
       );
       return;
     }
--- a/src/components/pages/Onboarding/InputInviteToken.tsx
+++ b/src/components/pages/Onboarding/InputInviteToken.tsx
@@ -98,7 +98,7 @@ class InputInviteTokenView extends React.Component<
                 </Text>
               )}
               <Button
-                title={`Confirm invite token`}
+                title="Confirm invite token"
                 disabled={this.state.inviteToken === undefined}
                 style={styles.button}
                 onPress={() => {
@@ -106,7 +106,7 @@ class InputInviteTokenView extends React.Component<
                 }}
               />
               <Button
-                title={`Continue without invite token`}
+                title="Continue without invite token"
                 disabled={this.state.inviteToken !== undefined}
                 style={styles.button}
                 onPress={this.props.onContinueWithoutToken}
--- a/src/components/pages/Onboarding/InputInviteToken.tsx
+++ b/src/components/pages/Onboarding/InputInviteToken.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { View } from "react-native";
+import { View, ViewStyle } from "react-native";
 
 import {
   Button,
@@ -19,6 +19,7 @@ import { colors } from "react-native-elements";
 
 type OwnProps = {
   onInputInviteToken: (inviteToken: string) => any;
+  onContinueWithoutToken: () => any;
   onBack: () => any;
 };
 
@@ -76,7 +77,8 @@ class InputInviteTokenView extends React.Component<
           <View style={styles.body}>
             <View style={styles.card}>
               <Text style={{ fontSize: 18 }}>
-                Please enter your invite token:
+                If you were invited by an existing Raha member, please enter the
+                invite token you received in your invite email:
               </Text>
               <TextInput
                 autoFocus={true}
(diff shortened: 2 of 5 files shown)


src/components/pages/LogIn.tsx
```diff
@@ -446,7 +446,7 @@ class LogInView extends React.Component<LogInProps, LogInState> {
       const redirectParams = this.props.navigation.getParam("redirectParams");
       this.props.navigation.navigate(
         redirectTo,
-        ...(redirectParams ? redirectParams : {})
+        redirectParams ? redirectParams : {}
```
osdiab: meh bummer the types don't catch this, guess there's some leakage somewhere along the line?

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -171,6 +173,10 @@ class OnboardingView extends React.Component<OnboardingProps, OnboardingState> {
       this.steps.push(prevState.step);
     }
 
+    if (prevProps.isLoggedIn && !this.props.isLoggedIn) {
+      this._redirectIfLoggedOut();
```
tinaroh: Should we return if this happens?

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -182,7 +188,18 @@ class OnboardingView extends React.Component<OnboardingProps, OnboardingState> {
       this.state.verifiedName &&
       this.state.emailAddress
     ) {
-      this._goToStep(this._validatedCameraStep());
+      this._goToStep(this._validatedInviteTokenStep());
```
tinaroh: Should this case be removed now? This was for skipping the identity video if the user has a valid video already

tinaroh:
It's a lot of text for a technically optional step...
Can we say: "Were you invited by someone?" in header text,
"No" "Yes"

and if they click yes show the text input?

src/components/pages/Onboarding/OnboardingCreateAccount.tsx
```diff
@@ -67,17 +67,23 @@ class OnboardingCreateAccountView extends React.Component<
     const statusType = status ? status.status : undefined;
     switch (statusType) {
       case ApiCallStatusType.STARTED:
-        return <Text>Creating account...</Text>;
+        return <Text style={styles.paragraph}>Creating account...</Text>;
```
tinaroh: Do we override the invite video when the user creates an identity video?

rahulgi: nope

src/components/pages/Onboarding/InputInviteToken.tsx
```diff
@@ -88,24 +90,37 @@ class InputInviteTokenView extends React.Component<
...
+                style={styles.input}
               />
               {this.state.errorMessage && (
                 <Text style={[{ color: colors.error }, sharedStyles.paragraph]}>
                   {this.state.errorMessage}
                 </Text>
               )}
               <Button
-                title={`Confirm`}
+                title={`Confirm invite token`}
```
markulrich:
```suggestion
                title="Confirm invite token"
```

src/components/pages/Onboarding/Onboarding.tsx
```diff
@@ -277,28 +294,30 @@ class OnboardingView extends React.Component<OnboardingProps, OnboardingState> {
...
-      ? // The new member does not need to take a verification video
-        // if they have a valid joint invite video.
-        OnboardingStep.CREATE_ACCOUNT
-      : OnboardingStep.CAMERA;
+  _validatedInviteTokenStep() {
+    return this.props.hasValidInviteToken && this.state.inviteVideoIsValid
+      ? // The new member is not requested to input an invite token
+        // if they already have a valid one.
+        // Note, Previously we would send the user directly to CREATE_ACCOUNT
+        // if the onboarding video was joint. Now we require an identity video from
```
markulrich: I do like the idea of everyone having an identity video, but I also don't want to add it as a mandatory extra step for someone who already made a video with their friend as that feels very heavy for this early stage ... would be nice to let people see what Raha is about ASAP and then allow and encourage members to update their identity video at any time.

src/components/pages/LogIn.tsx
rahulgi: I think the leakage is in React-Navigation `getParam` returning `any` type. XD

src/components/pages/Onboarding/Onboarding.tsx
rahulgi: sounds like a good idea!

src/components/pages/Onboarding/Onboarding.tsx
rahulgi: let's give it a shot and can easily roll back if it turns out to be a sticking point

src/components/pages/Onboarding/InputInviteToken.tsx
rahulgi: 👍

src/components/pages/Onboarding/Onboarding.tsx
rahulgi: This is still used to check whether the video for the invite token that the user input exists before advancing.

PR #487, opened 2018-11-05
Move Mint operation types to raha-api

tinaroh:


src/store/selectors/activities/index.ts
```diff
@@ -5,14 +5,14 @@ import {
...
 } from "@raha/api-shared/dist/models/Operation";
 
 import {
   Activity,
   NewMemberRelatedOperations,
-  MintBasicIncomeOperation,
-  MintReferralBonusOperation,
   ActivityType,
   IndependentOperation
 } from "./types";
```
markulrich: Can we delete `MintBasicIncomeOperation`, `MintReferralBonusOperation`, and others from `.types`?

PR #495, opened 2018-11-09
Bundle tip operations at activity-level

tinaroh:
[1/3] for tipping #460 

- Tip operations are bundled as `childrenOps` on an Activity. I keep all operations in a map and do two passes so that if we decide to change the order of operations processed, it won't break
- Tips are combined into a TIP actor CallToAction
- Condense GIVE Stories to just say "donated to Raha Basic Income" inline instead of having 3 circles

### DIFF
--- a/src/components/shared/Story/index.tsx
+++ b/src/components/shared/Story/index.tsx
@@ -38,7 +38,8 @@ import { RahaState } from "../../../store";
 import { getLoggedInMember } from "../../../store/selectors/authentication";
 
 /**
- * TODO: test this for proper output
+ * Component for call-to-actions that allow the user to interact with a rendered
+ * story.
  */
 const CallToAction: React.StatelessComponent<{
   member: Member;
@@ -238,14 +239,14 @@ class ActivityContent extends React.Component<{
             }
           />
           {/*
-            * Everything in the description must ultimately be Text elements, or
-            * else React Native doesn't support it. This is done to ensure text
-            * flows correctly in descriptions.
-            */}
+           * Everything in the description must ultimately be Text elements, or
+           * else React Native doesn't support it. This is done to ensure text
+           * flows correctly in descriptions.
+           */}
           <Text style={styles.description}>
             {/*
-              * Name at most the first three actors, and just summarize the rest
-              */}
+             * Name at most the first three actors, and just summarize the rest
+             */}
             {actorsData === RAHA_BASIC_INCOME_MEMBER ? (
               <MemberName member={actorsData} />
             ) : (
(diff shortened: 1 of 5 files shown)


src/store/selectors/activities/index.ts
```diff
@@ -496,9 +556,31 @@ function addOperationToActivitiesList(
...
+              `Unexpected GIVE operation data type. Operation: ${JSON.stringify(
+                operation,
+                null,
+                2
+              )}`
+            );
+        }
+      } else {
+        // TODO(tina): Migrate GIVE operations to have metadata block
+        return addIndependentOperation(existingData, operation);
```
rahulgi: should this be `addDirectGiveOperation` instead?

tinaroh: no, addDirectGiveOperation requires the metadata block

src/store/selectors/activities/index.ts
```diff
@@ -453,6 +475,44 @@ function addResolveFlagMemberOperation(
...
+ * of activities to attach them.
+ */
+function addTipOperation(
+  existingData: ActivityBundlingData,
+  operation: TipGiveOperation
+): ActivityBundlingData {
+  const { activities, bundlingCache } = existingData;
+  const target_operation_id = operation.data.metadata.target_operation;
+  if (!target_operation_id) {
+    throw new Error(
```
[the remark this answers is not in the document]
tinaroh: This really should never happen as long as we have validation server-side. Also, I'd rather catch the issue early and leave it as an error until we have client-side logging that gets sent to the server.

src/store/selectors/activities/types.ts
```diff
@@ -32,10 +32,12 @@ export enum ActivityType {
  */
 export interface ActivityDefinition<
   Type extends ActivityType,
-  RelatedOps extends Operation | Operation[]
+  RelatedOps extends Operation | Operation[],
+  RelatedChildOp extends ChildOperation = never
```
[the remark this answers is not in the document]
rahulgi: oooh cool

src/components/shared/Story/index.tsx
```diff
@@ -265,8 +328,13 @@ class ActivityContent extends React.Component<{
         </View>
         {/* TODO(tina): Render call to action for tipping
         {actorCallToAction &&
+<<<<<<< HEAD
         actorsData !== RAHA_BASIC_INCOME_MEMBER && ( 
+=======
+        actorsData !== RAHA_BASIC_INCOME_MEMBER && ( // TODOt
+>>>>>>> 58687245... Add tip UI
```
rahulgi: merge conflict?

src/components/shared/Story/index.tsx
```diff
@@ -58,8 +68,61 @@ const CallToAction: React.StatelessComponent<{
               </TextLink>
             );
           case CallToActionDataType.TIP:
-            // TODO(tina): Make real component
-            return <Text key={idx}>{piece.data.tipTotal} as tip!</Text>;
+            return (
+              <View
```
rahulgi: maybe worth putting into it's own component?

src/store/selectors/activities/index.ts
rahulgi: oh

src/components/shared/Story/index.tsx
tinaroh: I'll do this in a later PR

src/components/shared/Story/index.tsx
tinaroh: ah oops i wasn't supposed to merge the UI changes into this PR

tinaroh: just fixed the bad merge, it happened because I was switching back and forth between my UI branch and this one

PR #498, opened 2018-11-10
Add functionality for navigating based on deeplink in notification.

rahulgi:


src/components/shared/Navigation.tsx
```diff
@@ -653,10 +664,35 @@ class NavigationView extends React.Component<Props> {
...
   }
 
+  /**
+   * Takes an action based on a notification being opened.
+   */
+  handleNotificationOpen = async (
+    notificationOpen: NotificationOpen | undefined
+  ) => {
+    if (notificationOpen) {
+      console.log(notificationOpen);
```
tinaroh: Remove?

src/components/shared/Navigation.tsx
```diff
@@ -653,10 +664,35 @@ class NavigationView extends React.Component<Props> {
...
+    this.handleNotificationOpen(
+      await firebase.notifications().getInitialNotification()
+    );
   }
 
+  /**
+   * Takes an action based on a notification being opened.
+   */
+  handleNotificationOpen = async (
+    notificationOpen: NotificationOpen | undefined
```
tinaroh: What's the different between | undefined vs ?

osdiab:
No difference although it might be that when you say undefined it has to
explicitly be provided as undefined, I forget if that’s the case

src/components/shared/Navigation.tsx
rahulgi: oops

src/components/shared/Navigation.tsx
rahulgi: (i think omar's comment was intended to be a reply to this).

Yep - the type is the same, but using `?` allows you to call the method without any arguments, while this forces you to always provide the argument, even if it's `undefined`.

PR #510, opened 2018-11-14
Add tipping UI

### LINKED ISSUE #460: Support tipping

tinaroh:
This adds the tipping experience! The user is given 5 seconds to cancel tips that are accrued by tapping on the tip button.

I need to figure out why the first render of the pending tip component is so slow, but I will follow up in separate PRs. I also left some button code commented out so that I can follow up on the styling tomorrow after Jon's feedback.

screenshot_20181115-014655_raha test
screenshot_20181115-020906_raha test

Closes #460


src/store/selectors/stories/index.ts
```diff
@@ -690,19 +695,31 @@ function createVerifyMemberStory(
...
   const tipData: TipData = tips.reduce(
     (data, tip) => {
-      if (tip.data.to_uid === member.get("memberId")) {
+      if (tip.data.to_uid === toMemberId) {
         return {
-          tipTotal: data.tipTotal.plus(tip.data.donation_amount),
-          fromMemberIds: data.fromMemberIds.concat(tip.creator_uid),
-          toMemberId: data.toMemberId
+          ...data,
+          tipTotal: data.tipTotal.plus(tip.data.amount),
```
rahulgi: Could we make this the total tip amount including donated amount? I think similar to give notifications, it makes sense for the displayed amount to be total tipped.

tinaroh: Do you think I need to display donated amount separately somewhere then?

PR #520, opened 2018-11-27
Bump RN Firebase version to v5.1.1.

rahulgi:
This results in major dependency jumps for associated firebase
libraries (notably google-play-services base which goes from 15 to 16 
on Android and iOS Firebase Pods which go from 5.3 to 5.11).

This commit also adds the Firestore dependency.

I successfully built and tested the app on Android. Successfully built
for iOS but would appreciate if either of you, @tinaroh and @osdiab,
could try installing it on your iOS device and playing around with it
to make sure everything still looks good.


tinaroh:
I won't have my iPad for the next two days, sorry. @osdiab can you test for him?

osdiab:
Sure will do later today

osdiab:
Checked, it seems to work for logging in via text, uploading verification videos.

osdiab:
man, committing all these dependencies is so big lol

rahulgi:
excellent. yeah it's pretty crazy lol

PR #529, opened 2018-12-30
Animate minting

markulrich:


src/components/pages/Wallet.tsx
```diff
@@ -310,17 +430,61 @@ const mapStateToProps: MapStateToProps<
...
+  OwnProps,
+  MergedProps
+> = (stateProps, dispatchProps, ownProps) => {
+  const {
+    loggedInMember,
+    mintableBasicIncome,
+    mintableInvitedBonus
+  } = stateProps;
+  var mintActions = [] as MintArgs[];
+  if (mintableBasicIncome && mintableBasicIncome.gt(0)) {
```
rahulgi: nit: given the typing as `Big` instead of `Big || undefined`, I think we can get rid of the existence check

